PR #7, opened 2020-08-29
Use submodules for nodesoup and CImg dependencies

lacc97:
Incorporate these dependencies into the tree as git submodules

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "nodesoup"]
+	path = source/matplot/3rd_party/nodesoup
+	url = https://github.com/olvb/nodesoup.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "nodesoup"]
 	path = source/matplot/3rd_party/nodesoup
 	url = https://github.com/olvb/nodesoup.git
+[submodule "cimg"]
+	path = source/matplot/3rd_party/cimg
+	url = https://github.com/dtschump/CImg.git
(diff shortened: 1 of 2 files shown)


alandefreitas:
Pardon my ignorance, but what are the cons of CPM compared to git submodules? CPM seemed much cleaner to me.

lacc97:
The main thing is that CPM is part of the build system instead of the version control system.
 - submodules are easier to work with on network-restricted build environments
 - submodules are easier to keep track of since everything is in the `.gitmodules` file
 - submodules are initialised into the source tree so it's easier to see what you've checked out
 - submodules are part of git so you get some integration with Github on other tools in the git ecosystem

The con of submodules is that they are somewhat more complicated to use. A good overview of submodules can be found [here](https://git-scm.com/book/en/v2/Git-Tools-Submodules).

lacc97:
I suppose in the end it's a matter of preference. Since it's your project, however, you have the final say.

damiandixon:
CPM is no longer actively maintained. Have to admit git submodules are more complicated to use.

alandefreitas:
@lacc97 Thanks. Those are the pros and cons that came to mind at first.

@damiandixon The last commits on the CPM repository were just 10 days ago so it should be ok. 

In any case, CPM is only a convenient wrapper around FetchContent, so it's not something to worry about. There are too many alternatives. In the end, it doesn't seem to make much of a difference so I guess I'll remain neutral on that for now.

lacc97:
Considering CImg is a single-header library, wouldn't it be better to just directly copy CImg.h into the tree somewhere instead of using CPM?

alandefreitas:
I also thought about that. Also for nodesoup, from which I just need two functions and the original repository is not being maintained anyway. I was just waiting to see what other people think. Does that sound good?

@damiandixon Any thoughts?

PR #9, opened 2020-08-29
Fix compilation errors

mrexodia:
This compiles with LLVM 10 for Windows and the MSVC standard library:

`cmake -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang ..`

Also includes fixes for gcc 8 and some CMake modernization supersedes #8

PR #13, opened 2020-08-29
Fix more compile errors

mrexodia:
Follow up for #9 that I forgot to push

Tested on Ubuntu 18.04 with:

- clang 10
- gcc 8
- gcc 10

I think the README should be updated with some better instructions as to which packages to install, I just kinda randomly installed things until it compiled but I'm not sure if I'm getting a full feature set now or not...


alandefreitas:
That's great. Are you being able to run the examples?
I had to return with `set(CMAKE_CXX_STANDARD 17)` to make your last PR work here. I hope that's fine.
I'll update the README in the next days.

PR #15, opened 2020-08-30
Fix compilation for MSVC

mrexodia:
Most of the fixes are workarounds for weird compiler bugs/quirks in MSVC, but everything is still standard C++ and compile with clang and gcc just fine.

Thanks to @can1357 for helping me diagnose and fix some of the issues.

The main issue is that there are free functions that have the same name as the class they construct (`legend` for example). This is all fine, until the free function is a template. The fix is to put the template function in an anonymous namespace. I think this is a compiler bug in MSVC, but it actually becomes impossible to call the class constructor if there is a function with that name available.

Another issue is that `typename... Args` is being abused. In many places there is only a single overload and the free function should just have the arguments explicitly instead of forwarding them. This helps with both readability and intellisense when fixed. I only changed `line`, which was problematic because of the compiler bug.

Other issues are with the pattern `class X` being used in many places where it's not necessary, this seems to trip up MSVC and cause internal compiler errors when done in a template. The `class X` should in my opinion only be used to disambiguate between a function and class, but again I only fixed the ones that crashed MSVC.

The world_map compilation speed improvement is necessary, because the initializer_list overload of the vector constructor is evaluated at compile-time and ultra slow. The fix shouldn't have any downsides.

Finally a lot of warnings come from truncating conversations from double -> float. I simply silenced them as a PUBLIC compile flag, instead of changing `1.45` to `1.45f` in hundreds of locations. The library still compiles with a lot of warnings, but most of the examples are now warning-free.

### DIFF
--- a/examples/polar_plots/ezpolar/ezpolar_1.cpp
+++ b/examples/polar_plots/ezpolar/ezpolar_1.cpp
@@ -1,5 +1,4 @@
 #include <cmath>
-#include <cstdbool>
 #include <matplot/matplot.h>
 
 int main() {
--- a/source/matplot/CMakeLists.txt
+++ b/source/matplot/CMakeLists.txt
@@ -91,6 +91,12 @@ target_link_libraries(matplot PUBLIC nodesoup cimg std::filesystem)
 # https://cmake.org/cmake/help/v3.14/manual/cmake-compile-features.7.html#requiring-language-standards
 target_compile_features(matplot PUBLIC cxx_std_17)
 
+# Hacks to support MSVC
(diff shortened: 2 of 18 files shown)


alandefreitas:
MSVC always has these kinds of bugs.

The `typename... Args` was there on purpose. I didn't see the point in writing lots of functions that would do exactly the same. I was also worried about code analysis at first. But it ended up working fine in CLion.

image

The conversations from double to float can be fixed over time.

Thanks a lot for the great work.

xnorpx:
Great work thanks!

PR #17, opened 2020-08-30
Fix searching for glfw3

dimztimz:
You need to install glfw3 previously, e.g. sudo apt install libglfw3-dev.


alandefreitas:
Hi @dimztimz,

I understand that. If you install glfw3 on your system, then find_package(glfw3) works without FindGLFW3.cmake. I also understand FindGLFW3.cmake is going to be worse than using the original GLFW3Config.cmake.

The problem is that this is an optional dependency. So I don't want people who don't have GLFW3 installed to get an error. And it's very very optional. It's just there for one of the examples. There are some good reasons for not having any optional requirements at all but that's a debate for another time.

I see three options: 1) Including FindGLFW3.cmake as it right now, which is not good as you know, 2) Looking for a GLFW3Config.cmake before calling find_package(glfw3), or 3) creating a new option `option(BUILD_OPENGL_BACKEND "..." OFF)` so the default option makes everything clear and the user later understands why CMake is coming up with the error.

On another point, I noticed you removed `set(CMAKE_CXX_STANDARD 17)` from CMakeLists.txt. Is there a big reason for that? I don't know if `set(CMAKE_CXX_STANDARD 17)` is the best option (maybe it's best to set it for each target), but in any case, the library doesn't compile without it:

```
55:55:11: : 55: errorerror: : error: a space is required between consecutive right angle brackets (use '> >')a space is required between consecutive right angle brackets (use '> >')

a space is required between consecutive right angle brackets (use '> >')using adj_list_t = std::vector<std::vector<vertex_id_t>>;

                                                      ^~
                                                      > >
```

dimztimz:
I pushed a new commit which fixes your concerns. `find_package(... QUIET)` does not show errors.

alandefreitas:
Great! That does it!

KnIfER:
I downloaded glfw prebuit package `glfw-3.3.4.bin.WIN64.zip`, assign the extracted path to glfw3_DIR, but cmake can't find it.

alandefreitas:
@KnIfER,

When Cmake calls `find_package(glfw3 QUIET)`, it will look for GLFW3 in [these directories](https://cmake.org/cmake/help/latest/command/find_package.html#id5). This also assumes your glfw3 version has a config file or your project has a custom `FindGlfw.cmake` in your `CMAKE_MODULE_PATH`. 

I'm not sure glfw3 provides a config file yet. Matplot++ provides no custom `FindGlfw.cmake` but the build script will [download and build glfw3](https://github.com/alandefreitas/matplotplusplus/blob/69cbc2d5fa0a46c41f7d79c592453fa0e52d0e36/source/matplot/CMakeLists.txt#L238) at configure time if `BUILD_EXPERIMENTAL_OPENGL_BACKEND` is `ON` and glfw3 is not found. This shouldn't lead to ODR violations because, if glfw3 is cannot found elsewhere, it's relatively safe to assume you're not linking it because it doesn't exist elsewhere.

Issue #11, opened 2020-08-29
A library should not use its dependecies from source


dimztimz:
Instead, it should strictly use `find_package` or maybe `find_library`. There shouldn't be CPM.cmake, no Hunter, no `FetchContent` + `add_subdirectory`. Those things are useful only for executables. If you are bundling the dependencies, you can easily get ODR violations. Additionally you make the library harder for packaging in other package managers like apt or vcpkg. For example

```
My Prog ---- depends -----> matplot++ ----- depends ------> bundled/downloaded glfw
          \
           -----------------------------------------------> system glfw
```

And now you have a ODR violation because you have 2 glfw in the program. See this CppCon talk for more info https://www.youtube.com/watch?v=sBP17HQAQjk

As a middle ground, you should always use `find_package` for popular libraries and use CPM for stuff that isn't available in other package managers.

alandefreitas:
You're absolutely right. I'll try to come up with a FindGLFW.cmake and FindGLAD.cmake to solve this.

dimztimz:
For glfw you don't need FindGLFW.cmake, it provides all the necessary files (glfw3Config.cmake and others). You need only this

```cmake
find_package(glfw3) # notice the 3
target_link_libraries(.... glfw) # no 3 here
```

I don't know for the others.

alandefreitas:
That's great. For GLFW3, it would probably require some extra code to make sure CMAKE_MODULE_PATH has access to glfw3Config.cmake or FindGlfw3.cmake. Otherwise, users would get

```
By not providing "Findglfw3.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "glfw3", but
  CMake did not find one.
```

But that's not complicated though. I'll have a look at that. 

GLFW3 and GLAD are just dependencies for the example of what an OpenGL backend could look like. In the future, this should probably move to another repository anyway.

alandefreitas:
That should solve it
https://github.com/alandefreitas/matplotplusplus/commit/92bd21ec56cf27ec8d389121a878f1a90cff0a8d

dimztimz:
You should fix other dependencies too like cimg and nodesoup.

CImg is popular library already shipped in Linux distributions. For cimg, find_pacakge() can't work. You can use something like `find_path` to search if the header `CImg.h` exists.

alandefreitas:
[...]

PR #21, opened 2020-08-30
Add option to choose between bundled CImg and nodesoup or system versions

lacc97:
Related to #11.


alandefreitas:
That's great. Maybe, in the future, we could do the same for some of those image libraries in that same file.
I think some of them are just one or two files. That would solve a lot of the problem with optional libraries.

PR #25, opened 2020-08-31
Enable install with CMake

lacc97:
This PR enables installation with CMake.

The nodesoup and CImg headers are installed into `include/matplot/deps` when using the bundled dependencies. The nodesoup static library also gets installed (not sure if this is really necessary).

If installed to a system location, Matplot++ can be used in any CMake project with `find_package(Matplot++)` and linking with the `Matplot++::matplot` target.

The experimental OpenGL backend does not get installed.


alandefreitas:
That's just great

lpea:
Thank you for this, @lacc97!

> This PR enables installation with CMake.
> 
> The nodesoup and CImg headers are installed into `include/matplot/deps` when using the bundled dependencies. The nodesoup static library also gets installed (not sure if this is really necessary).
> 
> If installed to a system location, Matplot++ can be used in any CMake project with `find_package(Matplot++)` and linking with the `Matplot++::matplot` target.
> 
> The experimental OpenGL backend does not get installed.

It would be awesome to have this piece of information in the README. Also, if used as an external dependency in another project, CMake 3.15 is required as `list(POP_BACK ...)` was added in CMake 3.15, as mentioned in the [Changelog](https://cmake.org/cmake/help/v3.15/release/3.15.html#commands).

Kudos to @alandefreitas, this library seems great!

alandefreitas:
Hi @lpea,
Thanks for the comments. That's on the plans.
I don't really remember the cross-platform CMake commands to install the library by heart.
I'll have a look at that later, and test on a few systems to make sure I'm providing the correct information in the README.
Thanks,

Issue #27, opened 2020-09-01
When closing gnuplot window manually, gnuplot exit command is not run correctly (Windows)


xnorpx:
If I close the gnuplot command and then press enter in my commandline it seems the run command is executed this way. 

```
Press ENTER to continue...

gnuplot> e
         ^
         line 49: invalid command


gnuplot> xit
         ^
         line 49: invalid command
```

Adding a flush before run_command("exit") fixes this

bikash-jaiswal:
I am using Arch Linux
![image](https://user-images.githubusercontent.com/11890268/91794633-8fc23c00-ebe9-11ea-8d63-cfc3da32b9df.png)

I am getting sh: gnuplot: command not found error
fixed it by manual installing gnuplot  

and then after closing the gnuplot window manually, program doesn't exit.
![image](https://user-images.githubusercontent.com/11890268/91794506-378b3a00-ebe9-11ea-888b-9c3730793229.png)

alandefreitas:
@bikash-jaiswal Maybe this PR has solved your problem, but it's a little different from the problem @xnorpx had.

Matplotplusplus is using a [process pipe](https://man7.org/linux/man-pages/man3/popen.3.html) to communicate with Gnuplot, which itself is using a [QLocalSocket](https://doc.qt.io/qt-5/qlocalsocket.html) to communicate with a Qt Application that shows the window. The way QLocalSocket works depends on the operating system. If your default Gnuplot terminal is not `qt`, this might even be another kind of communication. 

Depending on how this interprocess communication is implemented, Gnuplot might not know the window is closed. It tries to run `exit` to close it and it returns an error. It then returns an error to Matplotplusplus. Because pipes are unidirectional, matplotplusplus can do nothing about it. There is a workaround to make the pipes "bidirectional" with two pipes but it's offers very limited options. It wouldn't solve this specific problem.

So it's best to manage the Gnuplot backend programmatically. This can be improved with some other backends. They could also make the experience more uniform across platforms. But that's another story. I'm currently implementing a backend based on [OpenGL](test/backends/main.cpp), but this is going to take a while.

PR #29, opened 2020-09-01
Fix split exit command on windows

xnorpx:
Fix for #27


alandefreitas:
Nice one. Now that I see it, I wonder why this *wasn't* a problem on other systems.
I forgot to mention you can `#define TRACE_GNUPLOT_COMMANDS` to see what commands it is sending to gnuplot on the background.

PR #34, opened 2020-09-02
Make main dependencies private

lacc97:
Nodesoup, CImg and std::filesystem dependencies are only required during build and are not part of the public header API. Therefore, they can be made private, and we don't have to install them.

Unfortunately, CMake complains about nodesoup when building Matplot++ as a static library so we must still install `libnodesoup.a` in those cases.


alandefreitas:
Thanks!

PR #36, opened 2020-09-03
Install private dependencies when building as static library

### LINKED ISSUE #35: Cimg linker error
**Describe the bug**
Unable to compile due to linker error.

**Steps to Reproduce**
1. Matplot++ folder is seperated from my code (no using add subdirectory)
2. make && make install matplot++ package.
3. Attempt to use matplot++ package on my code.

**Output**
```
/usr/bin/ld: cannot find -lcimg
```

**Environment:**
 - OS: 18.04
 - Compiler: g++ and clang++
 - Compiler version: 10

**Additional context**
Currently I am able to both `make` and `make install` matplot++. However, when I tried using it on my code, I got linking error regarding cimg. Previous I am able to compile my code by just having the following in my `CMakeLists.txt`. 
```
find_package(Matplot++)
find_package(Filesystem) 

add_executable(run main.cpp)
target_link_libraries(run PUBLIC Matplot++::matplot)
```

However, if I'm using similar CMakeLists.txt now, I'm having linker error of `cimg`. I think it might be related with [this](https://github.com/alandefreitas/matplotplusplus/commit/44e186e9e6916800d930f7b73a85d73223820cce). Thus, if possible, I would like to ask what else do I need to add in my CMakeLists.txt in order to not having this error.

lacc97:
Although cimg, nodesoup and std::filesystem are all private dependencies of Matplot++, they must still be installed when Matplot++ is built as a static library. Otherwise, library consumers get linker errors due to missing targets.

This should fix #35.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,4 +46,9 @@ if (MASTER_PROJECT)
         INSTALL_DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/Matplot++)
     install(FILES ${CMAKE_CURRENT_BINARY_DIR}/matplot++-config.cmake
         DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/Matplot++)
+
+    if(NOT BUILD_SHARED_LIBS)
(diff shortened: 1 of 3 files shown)


alandefreitas:
Thanks!

PR #40, opened 2020-09-04
Fix Linux CI

xnorpx:


alandefreitas:
Hi @xnorpx, 
Thanks! That's great. I was just about to start thinking about that.
I had a lot at [your solution](https://github.com/alandefreitas/matplotplusplus/actions/runs/240019511) and I noticed it doesn't even need the apt-get step. The step is returning:
```
g++-8 is already the newest version (8.4.0-1ubuntu1~18.04).
```
So I [tested](https://github.com/alandefreitas/matplotplusplus/runs/1073772292?check_suite_focus=true) the action again on another branch, and it seems like all we need is this bit:
```yml
args: "-DCMAKE_C_COMPILER=/usr/bin/gcc-8 -DCMAKE_CXX_COMPILER=/usr/bin/g++-8",
```
Who would have thought?

xnorpx:
ok cool, let me rebase and force push again

xnorpx:
@alandefreitas if you want to speed up the build maybe add option to just build small subset of the examples programs or just have one program that exercise all api. The linking for the individual .exe is what kills the build time.

alandefreitas:
> build maybe add option to just build small subset of

That's a good idea. I just have to think a little about what subset that would be.
The examples in [`appearance`](examples/appearance) must cover almost everything.
Coming up with a program that goes through the API is also not a bad idea, but that takes more time to write.

alandefreitas:
> ok cool, let me rebase and force push again

Amazing. So helpful. Thanks!

PR #45, opened 2020-09-05
Fix compile error caused by line_spec.cpp utf8 character.

sammi:
Fixed compiler error in issue: https://github.com/alandefreitas/matplotplusplus/issues/38


alandefreitas:
Great! Thanks!

PR #48, opened 2020-09-06
Fix perfect forwarding

lacc97:
Should be `Args&&... args` for parameters and `std::forward<Args>(args)...` for forwarding.


alandefreitas:
Makes sense. Thanks!

PR #50, opened 2020-09-08
Reduce possibly expensive copying and allocating

lacc97:
Modify the `to_vector{1,2,3}d` functions so that no copy is performed if passed a parameter of type `const vector_{1,2,3}d&`, instead the reference itself is forwarded and returned.

Use `std::string_view` in place of `const std::string&` in parameters to avoid copying and allocating when passing C-style strings. This is also more flexible in the sense that the user can pass any type (even a user defined type) that is convertible to a `std::string_view` while using a `const std::string&` requires always constructing an `std::string` (which is more expensive since it has to allocate memory and copy data). I haven't touched parameters of type `std::vector<std::string>`.

### DIFF
--- a/source/matplot/axes_objects/bars.cpp
+++ b/source/matplot/axes_objects/bars.cpp
@@ -88,7 +88,7 @@ namespace matplot {
         return res;
     }
 
-    std::string bars::legend_string(const std::string &title) {
+    std::string bars::legend_string(std::string_view title) {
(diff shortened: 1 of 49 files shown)


alandefreitas:
That's amazing. Thanks, man!



This is a very welcome PR. I had something like that in my mental to-do list.

PR #55, opened 2020-09-12
Rewrite contour generation

### LINKED ISSUE #52: About contour_plots
Hi, everyone,

I find the contour_plots/* have problems that the results are arbitrary. I have tested all of the others and the results are ok. Is there anybody find the contour_plots problem in this version.

Finite

lacc97:
This implementation is based on gnuplot's contour generation implementation. 
This also fixes #52.

Given the (x,y,z) points, we generate a mesh of edges and triangles for the (x,y) grid. To generate a contour for a given level, first we go through all edges and mark them if the edge crosses the given z-plane. Finally, we trace the contour line by picking a marked edge and following through adjacent triangles to other marked edges until we either hit the grid boundary or we close the contour line.

Filled contour generation is not implemented yet so this PR isn't ready for merging.

### DIFF
--- a/source/matplot/axes_objects/contours.cpp
+++ b/source/matplot/axes_objects/contours.cpp
@@ -21,16 +21,14 @@ namespace matplot {
         : axes_object(parent), X_data_(X), Y_data_(Y), Z_data_(Z),
           line_spec_(this, line_spec) {
         initialize_preprocessed_data();
-        contour_generator_ = QuadContourGenerator(X_data_, Y_data_, Z_data_,
-                                                  _corner_mask, nchunk_);
+        contour_generator_ = detail::contour_generator(X_data_, Y_data_, Z_data_);
     }
 
     contours::contours(class axes_type *parent, const vector_2d &Z,
                        std::string_view line_spec)
         : axes_object(parent), Z_data_(Z), line_spec_(this, line_spec) {
         initialize_preprocessed_data();
-        contour_generator_ = QuadContourGenerator(X_data_, Y_data_, Z_data_,
-                                                  _corner_mask, nchunk_);
+        contour_generator_ = detail::contour_generator(X_data_, Y_data_, Z_data_);
     }
 
     std::string contours::set_variables_string() {
--- a/source/matplot/axes_objects/contours.cpp
+++ b/source/matplot/axes_objects/contours.cpp
@@ -50,7 +50,8 @@ namespace matplot {
 
         std::stringstream ss;
         if (filled_) {
-            auto [lower_levels, upper_levels] = get_lowers_and_uppers();
+            std::vector<double> lower_levels, upper_levels;
+            std::tie(lower_levels, upper_levels) = get_lowers_and_uppers();
             // Command for background filled curve
             // The background polygon with the whole area has its level defined
             // by the largest polygon. Whatever level is outside this largest
@@ -61,6 +62,7 @@ namespace matplot {
 
             // We plot background only if one of the levels is below zmin
             bool plot_background = lower_levels[0] < zmin_;
+            std::array<float, 4> background_color;
             if (plot_background) {
                 auto &largest_segment = *filled_line_parents_[0];
                 auto line_index = largest_segment.line_index;
@@ -71,17 +73,18 @@ namespace matplot {
                     // background is 1 lower than parent on lower level
                     size_t level_index = line_index > 0 ? line_index - 1 : 0;
                     double background_z_level = lower_levels[level_index];
-                    line_spec_.color(parent_->colormap_interpolation(
+                    background_color = parent_->colormap_interpolation(
                         background_z_level, contour_min_level,
-                        contour_max_level));
+                        contour_max_level);
                 } else {
                     // background is 1 higher than parent on upper level
                     size_t level_index = line_index > 0 ? line_index - 1 : 0;
                     double background_z_level = upper_levels[level_index];
-                    line_spec_.color(parent_->colormap_interpolation(
+                    background_color = parent_->colormap_interpolation(
                         background_z_level, contour_min_level,
-                        contour_max_level));
+                        contour_max_level);
                 }
+                line_spec_.color(background_color);
 
                 std::string ls =
                     " '-' with filledcurve " +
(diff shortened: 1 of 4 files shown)


lacc97:
I am having a bit of trouble with the filled contour generation part. It's not very clear to me how the vertices list is structured or what is the purpose of the code list.

alandefreitas:
That's great. At the time, I didn't even consider the gnuplot implementation because it doesn't include filled contours but now that seems like a small problem compared to the issues with the matplotlib implementation. If the gnuplot implementation is working ok, we can already use it for the normal lines.

The thing with filled contours is that we need to identify which lines are holes inside other lines. Because we need to plot child lines last. We can sort the lines as a second step, which is not very complicated to do. But this can become very expensive and dominate the cost of generating contours. I don't know the asymptotic cost of doing that by heart, but I remember that it dominates the cost.

In any case, for now, it's probably better to use this expensive solution than to have the arbitrary lines we were getting from the matplotlib implementation. I haven't looked at the gnuplot implementation yet but I guess it also doesn't have the bad coding practices we see in matplotlib. It would take a while to fix that.

In the long run, however, filled contours need a small adaptation to the algorithm to identify holes as we generate the lines. It's a matter of changing the order in which we explore the space. Instead of looking for lines globally, we look for the next line only inside the polygon of the previous lines. This way, we can put the lines in order and don't need to sort them in the end.

In the matplotlib implementation, the codes are only there to indicate if we should create a line to the new point (the most usual code), start a new line at that point, or close a polygon at that point. It's perfectly possible to implement all that without these codes. It's just how matplotlib decided to implement it.

lacc97:
Thanks. I think I understand a bit better.

For the filled contours, I was thinking of using a procedure similar to the one for single line contours, except instead of marking each edge, we mark each triangle that intersects with the filled region. Then we go around generating the contour lines as usual except the marked triangles should tell us whether the line corresponds to a hole or not.

lacc97:
Actually now that I think about it, perhaps we could skip line generation entirely for filled contours and simply store the triangles themselves. These could then be rendered with calls to `backend::draw_triangles`. There's probably a gnuplot comand to do the same for the gnuplot backend. The only problem I see with this is increased memory usage of storing all the triangles and vertices.

alandefreitas:
I'm not sure I understand what it means to mark each triangle that intersects with the filled region. Is this a triangle inside each quad? These algorithms usually mark each quad as having an intersection or not and then create a point inside each intersecting quad by interpolating the 4 points around the quad. I'm not sure I understand where the triangles conceptually fit in there.

Isn't it possible to call the gnuplot function for contours only for a region of the space? If this is possible, we could call the function many times eliminating these regions. Dividing and conquering. Filled contours are usually generated for two levels at a time. If we find a line at level 1, we look for a line at level 2 in this subregion of the space. And vice-versa. That should work. I'm not sure.

lacc97:
I think I haven't properly explained how this algorithm works. Each quad in the grid is divided into two triangles and the corresponding edges for the triangles. Each triangle keeps track of its bounding edges and each edge keeps track of its two neighbouring triangles.

For the line contour generation, we first iterate over all edges and mark those that cross the z-level that we want (i.e. one of the edge vertices is above the z-level and the other is below). Then we iterate over all edges again, but upon finding

lacc97:
My idea for the filled contour, is to iterate over all triangles and mark those that intersect with the filled region (i.e. those where at least one of the vertices is in between the given z-levels). We can then directly draw these marked triangles and we get our filled region.

We could smooth it out a bit more  by splitting the boundary triangles into smaller triangles with vertices obtained by interpolating along an edge. It shouldn't be too difficult to find these boundary triangles as these are just triangles where one of the vertices is outside the filled region.

alandefreitas:
I see. So this is how the Gnuplot implementation currently works? I like the idea. It seems better than the quads. Are the triangles always (NW,NE,SW) and (SW,NE,SE)?

So your idea is to literally draw all (i) triangles that (ii) have edges (iii) through which the contour lines pass? I'm not sure I understand the idea correctly, but it seems like 

(i) this might be more difficult to implement and slower than just calling the other function repeatedly, sorting the contour lines, or adapt it to find lines in some appropriate order, 
(ii) it's probably going to be slower than plotting a single polygon and even slower with the Gnuplot process pipes (images, for instance, are very slow already), and 
(iii) it's going generate only an approximation of the line which is probably not worth it since we already have an algorithm to generate the "perfect" line.

I'm might be wrong on many of these points though. Because I'm not sure I completely understand the idea.

Maybe you can try just plotting the filled contour lines as they come from the Gnuplot algorithm. At first, they are probably going to be in the wrong order (maybe not?), but we can post process the vector of lines to put them in the right order. 

After generating the lines, sorting these lines might be expensive, but I think there are some logical constraints we can consider to make this faster. We could just call `std::sort` on the vector of lines, and use a lambda based on these contraints to put holes at the end. 

For instance, we can take their bounding rectangles (x_min, x_max, y_min, y_max) and use it to infer if one line can possibly be a hole of another line. If their bounding rectangles don't overlap, there's nothing to worry about. If they overlap, the smaller bounding rectangle must be the hole and should come last. 

I'm not sure, but my guess is that sorting these lines by their area, or the area of their bounding rectangles, should be enough to make sure we don't plot holes before "parent" lines. A hole, or even its bounding rectangle, shouldn't have a larger area than its parent.

lacc97:
> Are the triangles always (NW,NE,SW) and (SW,NE,SE)?
Triangles are always (NW, SW, SE) and (NW, SE, NE).

> So your idea is to literally draw all (i) triangles that (ii) have edges (iii) through which the contour lines pass? I'm not sure I understand the idea correctly, but it seems like
> 
> (i) this might be more difficult to implement and slower than just calling the other function repeatedly, sorting the contour lines, or adapt it to find lines in some appropriate order,
> (ii) it's probably going to be slower than plotting a single polygon and even slower with the Gnuplot process pipes (images, for instance, are very slow already), and
> (iii) it's going generate only an approximation of the line which is probably not worth it since we already have an algorithm to generate the "perfect" line.
Yes, this is exactly what I mean.
I don't think it should be especially slower or more difficult than the procedure for single line contours since it's essentially the same process (iterate over and mark edges vs iterate over and mark triangles).
I originally considered drawing triangles because I thought modern computers are supposed to be good at drawing triangles. But yes, I think you may be right about it being slower to draw especially on gnuplot.

> Maybe you can try just plotting the filled contour lines as they come from the Gnuplot algorithm. At first, they are probably going to be in the wrong order (maybe not?), but we can post process the vector of lines to put them in the right order.
> 
> After generating the lines, sorting these lines might be expensive, but I think there are some logical constraints we can consider to make this faster. We could just call std::sort on the vector of lines, and use a lambda based on these contraints to put holes at the end.
> 
> For instance, we can take their bounding rectangles (x_min, x_max, y_min, y_max) and use it to infer if one line can possibly be a hole of another line. If their bounding rectangles don't overlap, there's nothing to worry about. If they overlap, the smaller bounding rectangle must be the hole and should come last.
> 
> I'm not sure, but my guess is that sorting these lines by their area, or the area of their bounding rectangles, should be enough to make sure we don't plot holes before "parent" lines. A hole, or even its bounding rectangle, shouldn't have a larger area than its parent.
I think this works too. I'll try and implement this then.

alandefreitas:
Great. Now that I'm thinking about it, the texts I read about putting these lines in order might be overestimating this cost. That is unless this idea of sorting these lines by their area is wrong.

lacc97:
I think it's correct. Since contour lines shouldn't intersect, the bounding region of a child line should be strictly smaller than the parent line. If there is no parent-child relationship between two lines then, as you said, we don't have to worry about their relative ordering.

lacc97:
So an update on what I've been doing.

The filled contour generation part is almost complete. The outline of the process is:
1. Generate line contours (using the gnuplot algorithm) for the upper and lower values of z
2. Closed line contours don't need further processing at this point
3. Open line contours (i.e. contour lines that cross the region boundary) are used to partition the region into a bunch of closed polygons. From these, we pick those that are filled (i.e. not holes)

At this point we return from `contourc.cpp`, so the `contours.cpp` stuff only receives closed polygons to draw. This simplifies things for `contours.cpp`.

There are a few edge cases that I have to fix (e.g. when there's a hole in a corner but then another filled contour inside the hole) but beyond that it's mostly complete. In terms of performance it's not noticeably slower than the original implementation but I haven't rigorously measured it yet.

alandefreitas:
Thanks for the update. That's going to be so much better that the matplotlib implementation. It should probably be faster also.

The matplot implementation keeps higher values of z always on the left side of the line contour. Does the gnuplot implementation provide a way to know which side of the line contour should be colored? In any case, I think there's a function somewhere in contours.cpp that tests which side should be colored anyway.

If that's an option, we could use almost the exact same procedure from contourc.cpp and the exact same lines from contourc.cpp for filled contours because we could go around the boundaries to make them all closed polygons. 

In that case, we wouldn't even need pairs with upper and lower values of z. We would just need to plot these closed polygons ordered by size and we wouldn't need two sets of contour lines in filled contours anymore.

That sounds like what you already doing but I got a little confused when you said "upper and lower values of z". I haven't seen the code yet but I hope that's helpful.

Thanks again.

alandefreitas:
Hi @lacc97,

I've been thinking about this PR and, since https://github.com/alandefreitas/matplotplusplus/pull/81 already fixed the original problem, wouldn't it be better to just modernize the contourc algorithm instead of reimplementing it? By modernizing, I mostly mean just replacing macros with constexpr, and categorial ints with enums.

Besides, I suspect the gnuplot license is not compatible with the MIT license. I'm not completely sure about that. I'll look into it.

lacc97:
Hi @alandefreitas,
Sorry for just disappearing a couple of months, I've been busy with uni and haven't really had much time to code.

You're right, this PR has been superseded by #81, so I'll close it now. I may try to work on what you mentioned during holidays, but no promises though :\

Issue #61, opened 2020-09-16
Undefined reference in matplot.lib with Visual Studio 2017


PJ127:
**Bug category**

- [x ] *bug - compilation error*
- [ ] *bug - compilation warning*
- [ ] *bug - runtime error*
- [ ] *bug - runtime warning*
- [ ] *bug - logic error*

**Describe the bug**

```cpp
int main(int argc, char **argv) {
    using namespace matplot;
    std::vector<double> x = linspace(0, 2 * pi);
    std::vector<double> y = transform(x, [](auto x) { return sin(x); });
    plot(x, y, "-o");
    show();
    return 0;
}
```

I am trying to compile this project (which works fine with MinGW) with Visual Studio 2017 and I get the following link 'unresolved symbol' errors:
```
80>matplot.lib(network.obj) : error LNK2019: symbole externe non résolu "class std::vector<struct nodesoup::Point2D,class std::allocator<struct nodesoup::Point2D> > __cdecl nodesoup::fruchterman_reingold(class std::vector<class std::vector<unsigned __int64,class std::allocator<unsigned __int64> >,class std::allocator<class std::vector<unsigned __int64,class std::allocator<unsigned __int64> > > > const &,unsigned int,unsigned int,unsigned int,double,class std::function<void __cdecl(class std::vector<struct nodesoup::Point2D,class std::allocator<struct nodesoup::Point2D> > const &,int)>)" (?fruchterman_reingold@nodesoup@@YA?AV?$vector@UPoint2D@nodesoup@@V?$allocator@UPoint2D@nodesoup@@@std@@@std@@AEBV?$vector@V?$vector@_KV?$allocator@_K@std@@@std@@V?$allocator@V?$vector@_KV?$allocator@_K@std@@@std@@@2@@3@IIINV?$function@$$A6AXAEBV?$vector@UPoint2D@nodesoup@@V?$allocator@UPoint2D@nodesoup@@@std@@@std@@H@Z@3@@Z) référencé dans la fonction "protected: void __cdecl matplot::network::process_force_layout(void)" (?process_force_layout@network@matplot@@IEAAXXZ)
[...]

PR #66, opened 2020-09-28
Fix documentation on using system-installed Matplot++

lacc97:
May have contributed to issues #61 and #65.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2053,7 +2053,7 @@ If you are using CMake, you can then find Matplot++ with the usual `find_package
 ```cmake
 find_package(Matplot++ REQUIRED)
 # ...
-target_link_libraries(my_target PUBLIC matplot)
+target_link_libraries(my_target PUBLIC Matplot++::matplot)
 ```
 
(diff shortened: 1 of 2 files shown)


alandefreitas:
That's great. These PRs are *very* useful. It seems like more than half the issues are related to integration. There are so many ways in which people use the library. It's like an issue botleneck there. Thanks!

Maybe we could also name the internal target to `Matplot++::matplot` in https://github.com/alandefreitas/matplotplusplus/blob/7864cc34984362c70ce512bb8c0c5fa6586b3ca1/source/matplot/CMakeLists.txt#L3 to make names coherent for users want to use `add_subdirectory`? Or is it a bad idea? Not sure.

lacc97:
We can't name the actual target `Matplot++::matplot` (namespaces are only allowed for `IMPORTED` targets) but we can alias it:
```
add_library(Matplot++::matplot ALIAS matplot)
```

Could be something like
```
if(NOT MASTER_PROJECT)
    add_library(Matplot++::matplot ALIAS matplot)
endif()
```
to make sure it happens when Matplot++ gets included as a subdirectory.

alandefreitas:
Mmm... Interesting. Yes. That would work. Then the user can use either `matplot` or `Matplot++::matplot` in that case. It's even better: we can use `Matplot++::matplot` throughout the documentation and people who have been using `matplot` are still ok.

PR #83, opened 2020-10-17
add make uninstall target

### LINKED ISSUE #82: Add make uninstall target
**Feature category**
- [ ] *enhancement - build system*

**The problem**

After installing matplotplusplus there is no easy way to remove the installed files from my system.

**The solution I'd like**
Having a `make uninstall` target results in an easy way to remove the installed files.

N/A

**Additional context**

acxz:
Resolves #82


alandefreitas:
Hi @acxz,

That looks great. 

Could you add an extra step at the end of `matplotplusplus/.github/workflows/build.yml` just to make sure future PRs don't break your uninstaller?

It would be probably something like:

```yml
      - name: Uninstall
        working-directory: ./build
        run: ${{ matrix.config.sudocmd }} cmake --uninstall .
```

Or

```yml:
      - name: Uninstall
        if: matrix.config.os != 'windows-2019'
        working-directory: ./build
        run: ${{ matrix.config.sudocmd }} cmake --uninstall .
```

That would kind of serve as documentation as well.

acxz:
> Could you add an extra step at the end of matplotplusplus/.github/workflows/build.yml just to make sure future PRs don't break your uninstaller?

Done!

Since you mentioned docs, I also added a short blurb for the uninstall in the building from source section in the docs.

acxz:
Looking at the CI it seems like `cmake --uninstall .` doesn't work. I'll change it to the traditional `make uninstall`.

acxz:
Seems like a timeout on the macos build, gonna trigger the CI again.

alandefreitas:
Just had a look at `cmake/cmake_uninstall.cmake.in` and it seems like the command doesn't really depend on the make toolchain:

```cmake
exec_program(
       "@CMAKE_COMMAND@" ARGS "-E remove \"$ENV{DESTDIR}${file}\""
       OUTPUT_VARIABLE rm_out
       RETURN_VALUE rm_retval
       )
```

Am I right? I so, it might work on windows or anywhere else. This is actually more than a make only uninstall target.

We just have to run this custom target through cmake to uninstall. I don't remember the command by heart though.

Also, I had a look at the CI and it seems like, at least in the virtual machine, it's installing in `/usr/local/` and it's trying uninstall from `/home/runner/work/matplotplusplus/matplotplusplus/build/_CPack_Packages/Linux/TZ/` (Ubuntu) or `/Users/runner/work/matplotplusplus/matplotplusplus/build/_CPack_Packages/Darwin/TGZ/matplotplusplus-1.0.1-Darwin/` (Mac OS).

acxz:
> Am I right? I so, it might work on windows or anywhere else. This is actually more than a make only uninstall target.

I believe it should work on windows as well, which is why I was surprised that the CI failed for it.

hmm looking closer at the CI yeah it seems like something isn't working properly. On my machine the uninstall removes the files from /usr/local as it should, but I think something weird might be happening with CPack stuff. I'm afraid i'm not that familiar with CPack to figure out what should be done. I presume with CPack there are other methods of uninstalling?

With the simple:
```bash
mkdir build
cd build
cmake ..
sudo make install
sudo make uninstall
```
the pipeline works flawlessly.

alandefreitas:
Windows users don't usually use the make toolchain, so we need to call the install/uninstall targets through cmake. `sudo make install` doesn't generalize well.

In the CI, there must be something about this strategy of reading the manifest.txt that doesn't generalize well for all platforms. We'd need to see what's going on here.

I'm not really sure it has anything to do with cpack because cpack is kind of isolated from the cmake build script. All the cmake build script does is define some variables cpack might use when creating packages.

acxz:
> Windows users don't usually use the make toolchain, so we need to call the install/uninstall targets through cmake. sudo make install doesn't generalize well.

Yeah we should come up with a general solution.

> I'm not really sure it has anything to do with cpack ...

I see.

> In the CI, there must be something about this strategy of reading the manifest.txt that doesn't generalize well for all platforms. We'd need to see what's going on here.

So I took a glance at the CI, and I have a hunch. I think when we build the integration test and the CPack package the install_manifest.txt file gets overwritten which is what the `exec_command` above is trying to find. This means that the original install information is lost and `make uninstall` doesn't uninstall the files from the first install command.

This doesn't solve the general solution to get it to work on windows but if we just move the uninstall CI stage closer to original install instead of the Integration/CPack uninstall it should be good. I'll give it a shot.

acxz:
@alandefreitas sorry for the late follow-up.

> We just have to run this custom target through cmake to uninstall. I don't remember the command by heart though. Maybe cmake --build . --target uninstall?

Worked just like you expected!

Please let me know if there is anything else I should change before this PR can be merged.

alandefreitas:
That's great! Should work on all systems now. 👏

I've had a look at the CI script and the only minor detail is that you need to change the step order. 
Uninstalling should be the last step. You're uninstalling before the integration test so the following steps fail.
I changed the order in your branch. I hope that's ok. Have a look to check if it's fine.

It wasn't throwing an error before just because I allowed the integration test to fail. This is bad practice but I did it because 1) it was a new test, 2) this test is more like an example than a "real" test for now, and 3) CMake can't automatically find Matplot++Config.cmake on windows, or at least I don't know how to (PRs welcome here), so it always fails on windows for now.

acxz:
> So I took a glance at the CI, and I have a hunch. I think when we build the integration test and the CPack package the install_manifest.txt file gets overwritten which is what the exec_command above is trying to find. This means that the original install information is lost and make uninstall doesn't uninstall the files from the first install command.

> Uninstalling should be the last step. You're uninstalling before the integration test so the following steps fail.
I changed the order in your branch. I hope that's ok. Have a look to check if it's fine.

The reason why we had the uninstall target before the cpack command was because the `install_manifest.txt` gets overwritten.
I'm not sure of the best way to handle this properly besides, building, integration, uninstalling, building again, then cpacking.
Or if cpack also calls build then we don't have to build again, which I believe is what was happening before.

alandefreitas:
I see. Is the `install_manifest.txt` manifest created when we run cmake for the first time or after we run `cmake --install .`?

In any case, can't we save a copy of the original `install_manifest.txt` somewhere and ask the `uninstall` target to use this copy instead? In either case, we could call this copy the `uninstall_manifest.txt`.

acxz:
> I see. Is the install_manifest.txt manifest created when we run cmake for the first time or after we run cmake --install .?

Should be every time we run `--install`.

> In any case, can't we save a copy of the original install_manifest.txt somewhere and ask the uninstall target to use this copy instead? In either case, we could call this copy the uninstall_manifest.txt.

You would need to add some cmake logic to keep track of when you did a vanilla install and when you did a cpack install. I feel it would get complicated/not worth the effort just for CI. Mostly since I don't know how to do it off the top of my head.

I think placing the uninstall target after integration and before cpack is good enough.

alandefreitas:
> You would need to add some cmake logic to keep track of when you did a vanilla install and when you did a cpack install. 

We don't have to. We can just save the uninstall_manifest.txt anyway and always use regardless of cpack.

> I feel it would get complicated/not worth the effort just for CI. 

I'm actually thinking beyond the CI here. Because not many people will install the library and immediately uninstall it like in CI. People might want to run cpack for some reason between installing and uninstalling. Then the uninstall target will stop working and they will have no idea why. Then they open an issue. Most issues people open are related to how they followed the build script somewhat literally and it didn't work for some reason. 

> Mostly since I don't know how to do it off the top of my head.

`file(COPY <from> DESTINATION <to>)` to put it on the build dir or `install(FILES <from> DESTINATION <to>)` to install the manifest with other files. Depends on what you think makes more sense.

> I think placing the uninstall target after integration and before cpack is good enough.

That sounds like it solves the CI problem 👍. Much better. The other problem remains so we should at least document that it won't work if you run cpack between install and uninstall.

alandefreitas:
The idea of a complete solution seems to be stale and a parcial solution could lead to some extra problems so I'm closing this PR. Let me know if anything changes.

acxz:
sounds good

PR #88, opened 2020-10-18
add cmake 3.10 support

### LINKED ISSUE #87: Decrease minimum required CMake version
**Feature category**
- [x] *enhancement - build system*

**The problem**

While building for Ubuntu 18 works, users need to manually install their cmake versions to get things to build. As well as the default gcc version does not support fully c++17, in particular std::filesystem does not exist but instead there is std::experimental::filesystem.

Essentially my issue is that I would like to use matplotplusplus with the default versions of the dependencies available on Ubuntu 18.

**The solution I'd like**
Decrease the minimum cmake required version to 3.10 and allow the building with a slightly older gcc (gcc 7.5).

To get things working on older distributions where the gcc version is not 8 and the cmake version is not 3.14 there is not much that can be done, besides upgrading your distribution like going from Ubuntu 18 to Ubuntu 20 or manually installing new versions of these deps as already recommended.

**Additional context**

Related: https://github.com/alandefreitas/matplotplusplus/pull/59

acxz:
resolves #87


CMakeLists.txt
```diff
@@ -14,7 +14,9 @@ include(GNUInstallDirs)
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 message("CMAKE_MODULE_PATH=${CMAKE_MODULE_PATH}")
 option(CPM_USE_LOCAL_PACKAGES "Try `find_package` before downloading dependencies" ON)
-include(cmake/CPM.cmake)
+if (NOT ${CMAKE_VERSION} VERSION_LESS "3.14.0")
+    include(cmake/CPM.cmake)
```
alandefreitas, line 18: What happens with the AddCPMPackage commands when cmake_version < 3.14?
I can't see the results in continuous integration because their cmake_version is > 3.14. 
One of the problems with supporting older compilers, etc, is maintaining CI coherent with these promises.

acxz: Yeah that is functionality that would be lost for cmake users < 3.14. We don't have to guarantee that matplotplusplus will work on < 3.14, but for users that do try, they would be pleasantly surprised.

alandefreitas: That's fine. What I'm worried is that it might lose more than a functionality. 

If the cmake version < 3.14, It might just crash and not work at all when the script first calls AddCPMPackage. I don't know for sure because I haven't tested it yet (and that's kind of part of the problem) but the build script and the formulas for all package managers depend a lot on AddCPMPackage because it installs all the dependencies the user doesn't have.

If it's not going to work at all anyway, it's better to have an error message of the kind "please update cmake" than "the function AddCPMPackage does not exist". Cmake is usually so simple to update that we can even create an error message that tells the user what to type to update cmake on their system.

alandefreitas:
This seems to be stale and there doesn't seem to be a way to support older versions of CMake without removing important features from the build script or creating optional dependencies. Let me know if you want me to reopen that.

acxz:
I totally agree with you. I think having submitted the PR as well as made a discussions post about should be enough for users that want to install matplotplusplus under ubuntu 18 default cmake version.

At the very least we can point people to these resources when they open up potential issues about it.

CMakeLists.txt
acxz: > Cmake is usually so simple to update that we can even create an error message that tells the user what to type to update cmake on their system.

This may not always be the case for users that do not have admin privileges on their work machines (like me...)

But yeah I agree with you that, if we can't test it, we shouldn't merge it. Especially for build related commits.

PR #93, opened 2020-10-20
make bash build examples work

madronalabs:
Hi there, I'm building from source on OS X Catalina and contributing anything I'm finding that might be a hurdle. The bash scripts did not work as listed—possibly you have `mkdir foo` aliased to `mkdir foo; cd foo` on your system?

I had never used the -j option to cmake before—my feeling is that this might confuse unnecessarily and slow down builds on some machines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2319,7 +2319,7 @@ This will build the examples in the `build/examples` directory:
 mkdir build
 cd build
 cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="-O2"
-cmake --build . --config Release
+cmake --build . -j 2 --config Release
 ```
 
 On windows, replace `-O2` with `/O2`.
@@ -2332,7 +2332,7 @@ This will install Matplot++ on your system:
 mkdir build
 cd build
 cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="-O2" -DBUILD_EXAMPLES=OFF -DBUILD_TESTS=OFF 
-cmake --build . --config Release
+cmake --build . -j 2 --config Release
 cmake --install .
 ```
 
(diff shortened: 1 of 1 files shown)


alandefreitas:
Hi @madronalabs, thanks for the PR. Welcome.

You're right. The build script came from the CI script, and we ended up forgetting about `cd build`. Thank you for noting that. I'm pretty sure `mkdir build` is not aliased to `mkdir build; cd build;` anywhere. 😂 

Regarding `-j`, I don't see a reason not to use it or recommend not using it. It seems like a very useful feature to me, especially if you're building the examples. It saves a lot of time in continuous integration - it might save half an hour sometimes. All integration scripts, contributions, formulas for package managers, and IDEs I've seen so far use `-j`. Is there a reason why it would slow down builds on some machines? I've never seen anything like that, but if the feature does not generalize all, it seems like a good reason to remove it from the documentation.

On the other hand, it might be useful to tell people in the documentation to replace "2" with the number of cores they have available. Or even replace "2" with "4" or whatever is most common in the docs. 

It would be good if we could just find out the number of cores the user has for that so we don't need `-j`. If `-j` is not there, CMake will try to use [CMAKE_BUILD_PARALLEL_LEVEL](https://cmake.org/cmake/help/latest/manual/cmake.1.html) cores by default, but CMAKE_BUILD_PARALLEL_LEVEL is usually empty so it doesn't work. 

We could also replace `-j` with `--parallel`. With `--parallel` the intention is clearer and we don't have to explain as much.

Well, lots of options here.

madronalabs:
I've seen people alias `mkdir` to add the `cd` in the past, thus the wild guess on my part. 

Really bringing up the `-j` is such a minor detail. But it did make me go to the cmake docs to see what it was doing there because I was not familiar with it. Adding a comment would help that. I naïvely assumed cmake would use the maximum number of cores possible, in which case `-j 2` would often cause a slowdown.

alandefreitas:
So do you want to adjust the `-j 2` thing and still merge this PR?
I want to merge the `cd build` line.

madronalabs:
OK.

alandefreitas:
Great. Thanks. 👍

And welcome 😀:
* https://github.com/alandefreitas/matplotplusplus#contributors
* https://alandefreitas.github.io/matplotplusplus/contributing/contributors.html

Later we can discuss the `-j` thing. Maybe change it to `--parallel`.

PR #112, opened 2020-12-17
Fix NVCC compilation error

kaffehalv:
When compiling with NVCC a compilation error occurs

`_deps/matplot-src/source/matplot/core/axes_type.h:2790:38: error: ‘colormap_palette’ has not been declared 2790 | std::vector<std::vector<double>> colormap_{palette::default_map()}; `

that really is an error of NVCC, not the code itself, as it compiles just fine with g++. This fixes the error when using NVCC.

### DIFF
--- a/source/matplot/core/axes_type.h
+++ b/source/matplot/core/axes_type.h
@@ -2789,7 +2789,7 @@ namespace matplot {
             default_color(3), default_color(4), default_color(5),
             default_color(6)};
         size_t colororder_index_{0};
-        std::vector<std::vector<double>> colormap_{palette::default_map()};
+        std::vector<std::vector<double>> colormap_ = {palette::default_map()};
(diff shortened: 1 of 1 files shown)


alandefreitas:
Great. Thanks. And welcome.  🥳

I'll try to take that into account in the future.

Any idea why 

```cpp
std::vector<std::vector<double>> colormap_{palette::default_map()};
```

doesn't work but 

```cpp
size_t colororder_index_{0};
```

and others work?

GCC, Clang, and MSVC can handle both, so that's ok. I'm just curious so I can try to avoid any problematic syntax.

kaffehalv:
Thanks! :D I don't know, it is really weird. I've had trouble compiling other stuff with NVCC, but then the numbers of compilation errors where massive. So I didn't have time to try to find out why. But here it was only a single line.

PR #129, opened 2021-03-22
Fix float conversion warnings.

avocadoboi:
I only added a couple of static_casts and f literal suffixes to avoid conversion warnings. You should add more warning flags.


alandefreitas:
Looks good. Thanks! :D

alandefreitas:
The pedantic flags are [here](https://github.com/alandefreitas/matplotplusplus/blob/4a679e2ec0e0a6287c8273679fbbebd6871527ae/source/matplot/CMakeLists.txt#L142). Feel free to add more warning flags.

PR #133, opened 2021-04-10
Fix experimental filesystem support

myd7349:
Hi! I was trying to add this awesome library to `vcpkg` recently: https://github.com/microsoft/vcpkg/pull/13725. And the CI/CD system of `vcpkg` complains that it can not build this library on `x64-linux`. The build log can be seen here: https://dev.azure.com/vcpkg/public/_build/results?buildId=51823&view=results

According to those comments at the beginning of `FindFilesystem.cmake`, if no components are provided when calling `find_package`, it behaves as if the `Final` component was specified.

Only if we specify both of `Experimental` and `Final` components, it will falls back to experimental filesystem if final one is unavailable.

This will fix the build errors with g++ v7.

### DIFF
--- a/source/matplot/CMakeLists.txt
+++ b/source/matplot/CMakeLists.txt
@@ -1,4 +1,4 @@
-find_package(Filesystem REQUIRED)
+find_package(Filesystem REQUIRED COMPONENTS Experimental Final)
 
 add_library(matplot
         matplot.h


alandefreitas:
Hi @myd7349 !

That's great! Thanks!

PR #135, opened 2021-04-12
Reset axis_type directions when clearing axes_type

Zengit123:
The plot can inherit properties of a heat map and become reversed in Y-axis and have the minor_grid enabled as proposed in the issue #110 . Adding these 2 lines to each plot function solves the issue, but it is a very bad fix in my opinion. Ideally, every class should have a feature where its default settings are reset, otherwise this issue will appear forever.
I propose adding this as a temporary fix, close the #110 issue and create a new issue about objects not resetting their properties upon creation.

### DIFF
--- a/source/matplot/core/axes_type.cpp
+++ b/source/matplot/core/axes_type.cpp
@@ -47,7 +47,6 @@ namespace matplot {
         if (next_plot_replace_) {
             children_.clear();
             colororder_index_ = 0;
-            axes_type::clear();
         }
         children_.push_back(obj);
         touch();
--- a/source/matplot/core/axes_type.cpp
+++ b/source/matplot/core/axes_type.cpp
@@ -2318,6 +2318,8 @@ namespace matplot {
     line_handle axes_type::plot(const std::vector<double> &x,
                                 const std::vector<double> &y,
                                 std::string_view line_spec) {
+        this->y_axis().reverse(false);
+        this->minor_grid(false);
         axes_silencer s{this};
         line_handle l = std::make_shared<class line>(this, x, y, line_spec);
         this->emplace_object(l);
@@ -2326,6 +2328,8 @@ namespace matplot {
 
     line_handle axes_type::plot(const std::vector<double> &y,
                                 std::string_view line_spec) {
+        this->y_axis().reverse(false);
+        this->minor_grid(false);
         axes_silencer s{this};
         line_handle l = std::make_shared<class line>(this, y, line_spec);
         this->emplace_object(l);
--- a/source/matplot/core/axes_type.cpp
+++ b/source/matplot/core/axes_type.cpp
@@ -47,6 +47,7 @@ namespace matplot {
         if (next_plot_replace_) {
             children_.clear();
             colororder_index_ = 0;
+            axes_type::clear();
         }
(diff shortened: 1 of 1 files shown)


alandefreitas:
@Zengit123 Thanks for the PR! 

I'm just replicating my comment on #110 for convenience.

> The problem is clearer now.
> 
> When there are two different plot types (`axes_object`) in the axes (`axes_type`), the current heuristic is usually to set the best properties in the `axes_type` for the first plot or the latest plot we put there.
> 
> I don't think the current heuristic bothers to reset the properties of the `axes_type` when its list of `axes_object`s becomes empty.
> 
> This is what seems like a general solution:
> 1. If `gca()->clear()` between the commands that switch does not solve the problem, we need to change that to make sure `axes_type::clear` is clearing the relevant properties in this case.
> 2. We need to make sure `gca()->clear()` gets called when the list of `axes_object`s becomes empty in `axes_type::emplace_object`, to ensure consistent behavior when switching between plot types.
> 
> A less general solution, like editing the `axes_type::plot` or `axes_type::heatmap` directly, would only solve the problem for this very specific combination of plot types being switched.
>

Zengit123:
I have tried the clear functions you mention, and none of them worked. Interestingly enough, I tried to grep for "clear" in the whole matplot directory, couldn't find any relevant definitions. I did find other "gca()->" commands, but not the "clear" one.
Where is this method defined?

alandefreitas:
It's declared here

https://github.com/alandefreitas/matplotplusplus/blob/36d8dc6c3b94b7a71c4f129763f2c6ad8fc0b54a/source/matplot/core/axes_type.h#L594

and defined here

https://github.com/alandefreitas/matplotplusplus/blob/36d8dc6c3b94b7a71c4f129763f2c6ad8fc0b54a/source/matplot/core/axes_type.cpp#L2209

If you are using matplot++ with free-standing function, as `matplot::gca()` gives you a handle / shared pointer to the current axes_type in the current figure

https://github.com/alandefreitas/matplotplusplus/blob/36d8dc6c3b94b7a71c4f129763f2c6ad8fc0b54a/source/matplot/freestanding/axes_functions.h#L230

then calling `matplot::gca()->clear()` between the plots should reset the axes_type object to its original state so that the problem doesn't occur for any combination of plots. But it's true that the `clear()` function doesn't reset all relevant members of an axes_type object yet. It might not even be desirable to reset properties unrelated to `plot_object`s, such as fonts.

To solve that specific problem, the `emplace_object` function

https://github.com/alandefreitas/matplotplusplus/blob/36d8dc6c3b94b7a71c4f129763f2c6ad8fc0b54a/source/matplot/core/axes_type.h#L54

also needs to call `axes_type::clear()` at this point

https://github.com/alandefreitas/matplotplusplus/blob/36d8dc6c3b94b7a71c4f129763f2c6ad8fc0b54a/source/matplot/core/axes_type.cpp#L48

to make sure new child objects will not inherit the `axes_type` properties we created from the previous plot. The problem is simple it just didn't come up in our tests. Some other people probably had this problem but just adapted the properties they needed to make it look good.

alandefreitas:
That's great! Thanks! :D

PR #148, opened 2021-05-18
Quiver/Quiver3 Colormap and Normalize

matthew-hennefarth:
Currently quiver and quiver3 use the line_spec to define the color and style though it is very common to use a colormap to show the vector magnitude, and use normalized vector lengths. Furthermore, gnuplot has a completely different set of specification for [vectors and arrows](http://www.gnuplot.info/demo/arrowstyle.html) then what can be handled in the line_spec class (this is a completely separate issue which is not addressed in the PR). This change allows users to add a "magnitude list" (m-variable) that defines the color for the vector 

    quiver(x, y, u, v, m, scale, line_spec)
    quiver3(x, y, z, u, v, w, m, scale, line_spec)

Furthermore, a normalize(bool norm) function is added to the vectors object that allows the user to set all vectors to be the same length (actual length is dictated by the scale parameter). 

Note that these changes do not change the current examples at all as if an m list is not included, then it will instead define the user_color in the linespec as before. Only if the user passes the m vector does it switch to the gnuplot palette. An example usage is given in the examples/vector_fields/quiver3/quiver3_2.cpp file.


alandefreitas (commented):
Thanks for that! It's the best PR I've seen in a while!

I've sent a small PR in your fork trying to simplify the example a little. See what you think.
(Actually, it seems like I can't create a PR there from here but I can push changes. You can still rollback or change the commits as you like.)

The `vectors` class also includes a constructor for 2D fields and magnitudes. Do you have any simple example we could use for the `quiver` function? Something like a quiver_6.cpp. Unfortunately, this is how we "unit test" new functions, because we ultimately have to look at the plots and there's no easy way to automate that.

The examples are great. It's also worth including them in the README.md so that these nice images go to the documentation.

source/matplot/axes_objects/vectors.cpp
```diff
@@ -97,13 +142,24 @@ namespace matplot {
                     ss << "  " << 0.0;
                 }
 
+                if (normalize_) {
+                    double mag = sqrt(u_value * u_value + v_value * v_value +
```
alandefreitas, line 146: If I'm not wrong, I think some compilers require `std::sqrt` instead of just `sqrt`. I think the standard says these functions, when included from `<cmath>`, should be in the `std` namespace, but the calls outside `std` often end up working because some implementations just `#include <math.h>` inside `<cmath>` and they don't care to change the visibility of the original C functions.

alandefreitas (commented):
That's great. I'll have a look at the other example. I already ran `generate_examples` (our "unit tests") and everything seems fine. We just to need to push the images to the `docs` directory so that the README.md can find them. All the rest of the documentation is parsed from the README.md file with a [tool](https://github.com/alandefreitas/mdsplit) a wrote just for that.

source/matplot/axes_objects/vectors.cpp
matthew-hennefarth: Good to know, I will make the change:
```suggestion
                    double mag = std::sqrt(u_value * u_value + v_value * v_value +
```

source/matplot/axes_objects/vectors.cpp
```diff
@@ -97,13 +142,24 @@ namespace matplot {
                     ss << "  " << 0.0;
                 }
 
+                if (normalize_) {
+                    double mag = std::sqrt(u_value * u_value + v_value * v_value +
+                                      w_value * w_value);
+                    u_value *= scale_ / mag;
+                    v_value *= scale_ / mag;
+                    w_value *= scale_ / mag;
```
matthew-hennefarth, line 150: To prevent division by zero in the event u_value, v_value, and w_value are are zero (then vector points in no direction).
```suggestion
                   if (mag != 0.0) {
                        u_value *= scale_ / mag;
                        v_value *= scale_ / mag;
                        w_value *= scale_ / mag;
                    }
```

alandefreitas: Isn't the scale usually multiplied by the magnitude? I mean, shouldn't a larger magnitude make things larger regardless of the other normalization scale?

alandefreitas:
Great!

It seems like the links in README.md are broken though. The images in `docs/examples/vector_fields/...` are missing. The images are usually generated with `matplot::save` or just clicking the save button, depending on the gnuplot terminal, whatever is smaller. The thumbnail images are usually generated with `rsvg-convert`:

```bash
rsvg-convert -h 100 "<figure_destination_svg>" > "<figure_destination_thumbnail_png>";
```

That's what the `generate_examples` script does anyway, but any image of that size will do.

source/matplot/axes_objects/vectors.cpp
alandefreitas: Sorry. I just noticed this is inside an `if (normalize_)`. This is still seems weird though. If the ultimate effect of this normalization is just to attribute a constant to these values, isn't is easier just attribute this constant directly?

matthew-hennefarth:
Added the examples .svg and _thumb.png files to the docs/examples/vector_field directory.

source/matplot/axes_objects/vectors.cpp
alandefreitas: This seems like an edge case anyway, but maybe the ultimate intention is a unit_vector times `scale_` is the direction of `<u,v,w>` regardless of mag?

alandefreitas:
Great. I think it's only the `mag == 0` thing then. Maybe I'm just a little confused about that.

source/matplot/axes_objects/vectors.cpp
matthew-hennefarth: Yes it is supposed to be unit_vector*`scale_` in the direction of <u,v,w>. In the `axes_type::quiver` or `axes_type::quiver3` functions in the core/axes_type.cpp file, if scale is not 0, then u, v, and w are rescaled already:

    u_scaled = (u/u_max) * scale * x_max 

when we take the magnitude of <u_scaled,v_scaled,w_scaled> (which is <u,v,w> in the vectors.cpp), the scale gets implicitly pulled out and cancels with the scale above, and then all vectors go to unit length (length of 1). To adjust back to the scale that the user specified, we need to multiple again by scale. 

We could instead have normalize_ be a float (or double) instead of bool so that it would be

    float normalize_{0};
    float vectors::normalize() const { return normalize_; }
    class vectors &vectors::normalize(float normalize) {
        normalize_ = normalize;
        touch();
        return *this;
    }

Then in the plot_data() function it would multiple u,v,w by `*= normalize_/mag` instead.

matthew-hennefarth: There is also the issue that I just saw that if the user specifies `scale = 0`, then all the vectors will go to length 0 if normalized, when they should go to unit length.

alandefreitas: I forgot about the normalization in `axes_type::quiver` and `axes_type::quiver3`. It seems like a `float normalize_` would be redundant/ambiguous with `scale_`.

I'm worried the normalization in `axes_type::quiver` and `axes_type::quiver3` could make the current `vectors::normalize_` redundant or ambiguous? If so, it seems like a lazy normalization would be better for you. Instead of calculating unit vectors in `axes_type::quiver` and `axes_type::quiver3`, we could just set normalize_ to `true` in there, and you would later be free normalize it however you like.

In any case, this is an edge case and even any redundancy is hidden from the public API, so it's really up to you. I'm also fine with resolving this as it is if you'd like.

alandefreitas: > There is also the issue that I just saw that if the user specifies scale = 0, then all the vectors will go to length 0 if normalized, when they should go to unit length.

Beyond making this normalization / scales lazy, maybe it seems like a good idea to rename these things. For instance, there are many ways to "normalize" things. Maybe something like `bool unit_vectors_` would make what's happening clearer (unit vector * scale if true or direction * scale if false).

Also, `magnitudes` could just be `colors` or something like that, since this is an independent variable. A "magnitude" might make it seems like it's being inferred as `z=std::sqrt(std::pow(a,2)+std::pow(b,2)+...)` with zero memory cost instead of getting the values from another vector, which would provide the colors even if these don't match the vector magnitudes.

matthew-hennefarth: Ah, but they are not normalized in `axes_type::quiver` and `axes_type::quiver3` but just rescaled. So `vectors::normalize_` is not redundant (unless the user gives normalized vectors to begin with). 

Let me make the commit for making sure that if scale = 0 that we only divide by the mag instead since the user does not want to scale and only normalize vectors.

alandefreitas: I think part of the confusion is whether `*= scale_` should come before or after `normalization_` to unit vectors.

matthew-hennefarth: > Beyond making this normalization / scales lazy, maybe it seems like a good idea to rename these things. For instance, there are many ways to "normalize" things. Maybe something like `bool unit_vectors_` would make what's happening clearer (unit vector * scale if true or direction * scale if false).

So change `normalize_` to `unit_vectors_` and `normalize(bool)` to `unit_vectors(bool)`. Sounds good. 

> Also, `magnitudes` could just be `colors` or something like that, since this is an independent variable. A "magnitude" might make it seems like it's being inferred as `z=std::sqrt(std::pow(a,2)+std::pow(b,2)+...)` with zero memory cost instead of getting the values from another vector, which would provide the colors even if these don't match the vector magnitudes.

Sure, I will make that change. I will change magnitudes->colors and `m_data_` to `c_data_`.
 
> Vectors of length zero when scale = 0 seem like expected behavior.
It is only that in the `axes_type::quiver` and `axes_type::quiver3` if `scale = 0`, then you do not rescale the vectors at all, so it would be consistence with that.

alandefreitas: Great. I'm brainstorming here. Feel free to discard any of these suggestions.

matthew-hennefarth: > I think part of the confusion is whether `*= scale_` should come before or after `normalization_` to unit vectors.
> 
> It's probably more useful to have `scale_` applied after normalization so both can be used.

I am following the matplotlib convention here I guess. In python I would do

    q = ax.quiver(x, y, z, u, v, w, colors=c, length=0.25, normalize=True)

for a 3d vector field. So it sets the length of each vector to the same length, and then scales them so that each vector is only 0.25 units long.  So scale should occur after normalize otherwise there is no need to specify a scale parameter.

alandefreitas: > So scale should occur after normalize otherwise there is no need to specify a scale parameter

Exactly. Our problem is the scaling in `axes_type` is forcing the scaling to come before normalization.

> I am following the matplotlib convention here I guess. 

This function definition looks like a good API. 

- Your `magnitudes` become `colors` because they don't really need to match the magnitudes and the magnitudes are not coming from <u,v,w>. 
- The name `length` is also clearer than `scale`, because it make it clear that it will be applied even when normalization is `true`. The problem is this would need to happen lazily in `vectors` so that scaling could come after normalization. Or `normalize` will need to go inside `axes_type`, but this doesn't seems like a good idea because it wouldn't be able be undone later like the way you did initially.
- `normalize` and `length` are not ambiguous, which is nice. However, `length` would make it seem like it's *only* being applied when `normalize` is `true`, which is not currently true for matplotplusplus, or our API would break. So `scale` is probably the best name for us.
- `normalize` is also a reasonable name in this case, because it's not really going to be a unit vector after we change its length. I don't know if there's a better name for that.

source/matplot/axes_objects/vectors.cpp
```diff
@@ -147,9 +147,14 @@ namespace matplot {
...
                                   w_value * w_value);
                     if (mag != 0.0) {
-                        u_value *= scale_ / mag;
-                        v_value *= scale_ / mag;
-                        w_value *= scale_ / mag;
+                        u_value /= mag;
+                        v_value /= mag;
+                        w_value /= mag;
+                    }
+                    if (scale_ != 0.0) {
```
alandefreitas, line 154: We probably don't need `if (scale_ != 0.0)`. Although this is an edge case, if the user provided `scale` as `0.0`, they should be waiting for that. Although this might be inconvenient when the user is not expecting for that, this conditional could break the behavior when it's intentional. 

For instance, this could be part of an animation where the length is varying and passing through `0.0`. The scale could even the negative in this animation. In this case, whenever the length is 0.0, the user would see a large vector blinking, which is not expected behavior. 

Besides, although we are not dealing with that in matplotplusplus yet, more and more compilers are defining an error for floating point equality comparisons.

matthew-hennefarth: In `axes_type::quiver` and `axes_type::quiver3` if scale = 0, then no scaling is done. It is essentially saying turn off scaling if scale = 0 (treating it as a bool and a float). So currently if the user says scale = 0, they don't want to use scaling.

source/matplot/axes_objects/vectors.cpp
matthew-hennefarth: Okay I changed the `m_data_` and `std::vector<double> m` to `c_data_` and `std::vector<double> c` where appropriate, and changed comments to exclude magnitude and use color or color mapping. 

I agree with not putting normalize in the `axes_type` since it cannot be undone. It might be best then to place the scaling in the vectors class rather than in `axes_type::quiver`. Then the raw u,v,w data is stored in the class. Upon data_string() call we can then calculate u_scaled, v_scaled, w_scaled, or if normalize_ is true, scale the resulting vector by that amount. A possible vector constructor would be

    vectors(class axes_type *parent, const std::vector<double> &x_data,
                  const std::vector<double> &y_data,
                  const std::vector<double> &z_data,
                  const std::vector<double> &u_data,
                  const std::vector<double> &v_data,
                  const std::vector<double> &w_data,
                  const std::vector<double> &c_data,
                  double scale = 1.0, std::string_view line_spec = "");

In the `vectors::data_string()` method I would add something similar to (which is what is currently used)

         auto x_copy = x;
         std::sort(x_copy.begin(), x_copy.end());
         x_copy.resize(std::distance(x_copy.begin(),
                                     std::unique(x_copy.begin(), x_copy.end())));
         std::vector<double> x_diff(x_copy.size());
         std::adjacent_difference(x_copy.begin(), x_copy.end(), x_diff.begin());
         auto xdiffmin = std::min_element(x_diff.begin() + 1, x_diff.end());
         double u_max = *std::max_element(u.begin(), u.end());
          double x_max = xdiffmin != x_diff.end() ? *xdiffmin : u_max;

         auto y_copy = y;
         std::sort(y_copy.begin(), y_copy.end());
         y_copy.resize(std::distance(y_copy.begin(),
                                     std::unique(y_copy.begin(), y_copy.end())));
         std::vector<double> y_diff(y_copy.size());
         std::adjacent_difference(y_copy.begin(), y_copy.end(), y_diff.begin());
         auto ydiffmin = std::min_element(y_diff.begin() + 1, y_diff.end());
         double v_max = *std::max_element(v.begin(), v.end());
         double y_max = ydiffmin != y_diff.end() ? *ydiffmin : v_max;

         auto u_scaled =
             (scale != 0.)
                 ? transform(
                       u, [&](double u) { return (u / u_max) * scale * x_max; })
                 : u;
         auto v_scaled =
             (scale != 0.)
                 ? transform(
                       v, [&](double v) { return (v / v_max) * scale * y_max; })
                 : v;

So, if normalize_ is true the above code is not run, but instead the simple method of dividing by mag and mutliplying by scale is used. If normalize_ is false, then the above code is used instead. This also reduces code duplicity between axes_type::quiver and axes_type::quiver3

source/matplot/axes_objects/vectors.cpp
matthew-hennefarth: This is in axes_type::quiver
 
        auto u_scaled =
             (scale != 0.)
                 ? transform(
                       u, [&](double u) { return (u / u_max) * scale * x_max; })
                 : u;

source/matplot/axes_objects/vectors.cpp
matthew-hennefarth: We could make then the member variable scale_ of `vectors` an std::optional<double> so that if the user set scale_ it evaluates to true, and otherwise evaluates to false.

source/matplot/axes_objects/vectors.cpp
alandefreitas: I see. This was actually a workaround because there was no `normalize` yet 😆. 

You're absolutely right. Changing it would unfortunately break the current behavior. This seems like the wrong behavior but deprecating functions to change old behavior is quite problematic 🙁. 

So never mind. My bad. I'll try to remember about that when v2.0.0-alpha is out.

matthew-hennefarth:
At this point, I am good with the changes. The conversion of scale to an std::optional within the vectors class can occur in a different PR since it would involve changing some more of the internals.

source/matplot/axes_objects/vectors.cpp
alandefreitas: > It might be best then to place the scaling in the vectors class rather than in axes_type::quiver

That sounds great.

> A possible vector constructor would be

That looks good.

> So, if normalize_ is true the above code is not run, but instead the simple method of dividing by mag and mutliplying by scale is used

Sure. When `normalize` is false, we just change the lengths according to `scale`. When `normalize` is true, we do the same, but normalize before. We multiply by `scale` in both cases, like in our discussion about the python API. 

Regarding the normalization that comes before scaling, when `normalize` is `true`. If by `mag` you now mean the new `c_`, I don't think we need to divide the scales by the colors. Seems like bad semantics. If by `mag_` you mean the magnitude from <u,v,w>, i.e. making it a unit vector before scaling, then sure. I think you meant the later but just to be sure.

> We could make then the member variable scale_ of vectors an std::optional so that if the user set scale_ it evaluates to true, and otherwise evaluates to false

We can do that, but I think we don't really need to. This might require some changes in the API while `scale_ == 1.0` can achieve the same effect. If we want to be very pedantic about the processing cost, we could have a kind of [`if (!almost_equal(scale_, 1.0))`](https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon) before scaling, but I wouldn't worry much about that either. It seems like premature optimization all things considered.

alandefreitas:
> At this point, I am good with the changes. The conversion of scale to an std::optional within the vectors class can occur in a different PR since it would involve changing some more of the internals

Great. Workflow running then!

alandefreitas:
All checks have passed. Can I merge the PR?

matthew-hennefarth:
You can merge the PR!

alandefreitas:
Done! 😄 

Thanks for the PR! Best one in a while. I'll even create a new release soon, just to include it in `v1.1.0`.

I saw you have experience with vector fields. If you want/need to include anything more advanced in the library, feel free to open another PR. I might not completely understand all demands out there for vector fields.

When `v2.0.0` comes out, I'll port whatever is implemented here to the new backends. If everything goes as planned, it should support even cling and emscripten, so it'll very useful for scientific computing.

alandefreitas:
Documentation links (github actions takes a while):
https://alandefreitas.github.io/matplotplusplus/plot-types/vector-fields/quiver-3d/
https://alandefreitas.github.io/matplotplusplus/plot-types/vector-fields/quiver/

matthew-hennefarth:
Awesome, thanks! Yes, I currently do research in electric field in computational/theoretical chemistry so I work with visualizing these vector fields a lot. I have used matplotlib in the past but needed a c++ plotting library. My [CPET](https://github.com/matthew-hennefarth/CPET) program will be using this library now and I am hoping to submit a manuscript at some point in the next few months.

I plan to work on fixing the internals a little more for the vectors class so that it controls the scaling rather than having the `axes_type::quiver` or `axes_type::quiver3` handle it. I will also submit another PR regarding cmake and `-DBUILD_WITH_PEDANTIC_WARNINGS=ON` fairly soon.

I wish I could help out on the backend work that you seem to be doing, but alas, I have never worked with qt, opengl, or really any graphics backend before.

PR #149, opened 2021-05-19
Fix -DBUILD_WITH_PEDANTIC_WARNINGS=ON

matthew-hennefarth:
In the [contributing guidelines](https://alandefreitas.github.io/matplotplusplus/community/contributing-guidelines/) it specifies to build with `-DBUILD_WITH_PEDANTIC_WARNINGS=ON`. Using 

    cmake ../ -DBUILD_EXPERIMENTAL_OPENGL_BACKEND=ON -DBUILD_WITH_PEDANTIC_WARNINGS=ON
    make

results in the following [make_error.log](https://github.com/alandefreitas/matplotplusplus/files/6505326/make_error.log). The issue arises from CMake treating the header files from the `source/3rd_party` as source code and therefore warnings and errors can be applied when these headers are included in the matplot source code. A good solution has been to include the external library headers as system libraries (proposed on stack exchange [here](https://stackoverflow.com/questions/52135983/cmake-target-link-libraries-include-as-system-to-suppress-compiler-warnings). 

This PR adds the `target_link_libraries_system` cmake function which includes these libraries header files as system, which suppresses these warnings when building with `-DBUILD_WITH_PEDANTIC_WARNINGS=ON`. Then, all compiler warnings will only be generated for matplotplusplus code and not for any external libraries it uses or for any header files it includes. 

In the stack exchange post it was noted that this may not work for Visual Studio projects. It would be great to check this (I am unable to as I am on MacOS). Further it may have issues with transitive dependencies, but I do not believe this project has any.


matthew-hennefarth:
I will note that this may just be an issue with header files on my computer, though this would allow matplotplusplus to have more warnings if it is desired without having to worry about compatibility with external libraries.

alandefreitas:
Nice. Part of the problem is that I initially developed that on clang and later adjusted whatever warnings we had on GCC and MSVC. This left lots of small warnings in GCC and MSVC. Especially implicit conversion warnings in MSVC. For this reason, we are still not really able to be completely pedantic for all targets when `BUILD_WITH_PEDANTIC_WARNINGS` is `ON`. I'm still planning to push a single commit to make it completely pedantic and eliminate all these small warnings at once.

A second issue is, as you mention, bundled headers, like `CImg.h`, that have lots of deprecation errors. System libraries look like a good idea to solve the problem. We'd just need to check if this strategy to include header paths doesn't affect the generator expressions for include directories, which would make the cpack installers stop working. I don't think we will have any problem there but still need to test the installers to make sure.

The other idea I had was to just wrap these header inclusions with [explicit macros](https://nelkinda.com/blog/suppress-warnings-in-gcc-and-clang/) to disable and enable the relevant warnings:

```cpp
_Pragma("GCC diagnostic push")
_Pragma("GCC diagnostic ignored \"-Wunused-parameter\"")
#include <...>
_Pragma("GCC diagnostic pop")
```

The macros for MSVC are not that different. I'm not sure, but I imagine we wouldn't need many of these macros. The advantage is we wouldn't need to worry about the installers. The drawback is we would need to update that in case we find out about new warnings.

alandefreitas:
Looks like the bot finally put you in the documentation 😀

matthew-hennefarth:
Yes, I wasn't 100% confident that it would work with all the different ways matplotplusplus could be integrated into systems. Though it is completely possible as well to use an cmake if statement to change between the custome function `target_link_libraries_system` and the cmake function `target_link_libraries` if issues (or just define a new cmake build option).

alandefreitas:
> I'm still planning to push a single commit to make it completely pedantic and eliminate all these small warnings at once.

Done! 😀 https://github.com/alandefreitas/matplotplusplus/runs/2634535874?check_suite_focus=true

I'll create a release to include your vector fields and these fixes.

PR #159, opened 2021-06-07
Fixing MinGW issue due to  MSVC's "_dupenv_s" call

rath3t:
Hello everyone,
thank you for this nice library.

I tried to compile the library with mingw on windows and it worked but at linking in my project i witnessed the following error:

> source/matplot/libmatplot.a(gnuplot.cpp.obj):gnuplot.cpp:(.text+0x5921): undefined reference to `__imp__dupenv_s'

I fixed it with the following change since `_dupenv_s` is a Visual Studio only command and therefore mingw should follow the other branch.

I hope this is the correct way to do this PR. Thanks!

### DIFF
--- a/source/matplot/backend/gnuplot.cpp
+++ b/source/matplot/backend/gnuplot.cpp
@@ -37,7 +37,7 @@ namespace matplot::backend {
 
     gnuplot::gnuplot() {
         // 1st option: terminal in GNUTERM environment variable
-#if defined(_WIN32) || defined(_WIN64) || defined(__MINGW32__) ||              \
+#if defined(_MSC_VER) ||              \
     defined(__CYGWIN__)
         char *environment_terminal;
         size_t len;


alandefreitas:
Hi @rath3t,

Thanks for the PR! We had several issues and discussions about MinGW. 
I didn't even think the build script would work properly for MinGW without at least some small modifications.
All [checks](https://github.com/alandefreitas/matplotplusplus/actions/runs/915853920) have passed so that should work great.

If the build system is working ok for MinGW, it's probably worth including a MinGW config in our build workflow to make sure this is finally supported and ensure it keeps working with MinGW in future versions. 
It should just be a matter of including an extra configuration [here](https://github.com/alandefreitas/matplotplusplus/blob/b45015e2be88e3340b400f82637b603d733d45ce/.github/workflows/build.yml#L49) so we make sure this doesn't break in the future.

Moving this to another PR is also OK but it's probably worth doing it now so we already have the build script generating binaries for MinGW and ensure it never breaks. It's up to you though.

rath3t:
Thanks for the fast merge! I can try to add Mingw build there.
What bothers me even if i just write the Cmake statements inf 32 and 64 bit there with the correct minGW generators i don't know how the CI build and download some or the correct MinGW version. E.g. it worked at my pc with MinGW from https://winlibs.com/ where i used Win64 "GCC 11.1.0 + LLVM/Clang/LLD/LLDB 12.0.0 + MinGW-w64 9.0.0 - release 1" where i compiled with GCC 11.1.0. But i'm not sure what the CI does it the background. Therefore this needs to be also configured somewhere shouldn't it?

alandefreitas:
The virtual machine probably comes with MinGW installed. It's usually just a matter of copying+pasting one of the other configs, finding the correct path to MinGW, and giving it to Cmake in the config command. You don't even need the path sometimes, like [here](https://github.com/alandefreitas/matplotplusplus/blob/b45015e2be88e3340b400f82637b603d733d45ce/.github/workflows/build.yml#L45).

This [repository](https://github.com/actions/virtual-environments) has a list of included software for each platform.
The windows machine [comes with Mingw-w64 8.1.0](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md), which should be good enough I think.

There are probably some samples using MinGW in the repositories of some large libraries, but even large libraries often just test Clang, GCC, and MSVC ([example](https://github.com/ericniebler/range-v3/blob/master/.github/workflows/range-v3-ci.yml)).

PR #161, opened 2021-06-10
add MinGW workflow support

rath3t:
This PR adds a MinGW config to also test the MinGW compatibility. Hopefully this is enough to get it working.

Maybe i have to also specify the paths `-DCMAKE_C_COMPILER,  -DCMAKE_CXX_COMPILER`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
         config:
           # see: https://github.com/actions/virtual-environments
           - {
-            name: "Windows/2019/Static/X86/Release",
+            name: "Windows-MSVC/2019/Static/X86/Release",
             os: windows-2019,
             config: Release,
             cmake_extra_args: -G "Visual Studio 16 2019" -A Win32,
@@ -39,10 +39,19 @@ jobs:
             cores: 2,
           }
           - {
-            name: "Windows/2019/Static/X64/Release",
+            name: "Windows-MSVC/2019/Static/X64/Release",
             os: windows-2019,
             config: Release,
             cmake_extra_args: -G "Visual Studio 16 2019" -A x64 -DCMAKE_CXX_FLAGS="/O2",
+            sudocmd: "",
+            artifact_name: "Windows x64",
+            cores: 2,
+          }
+          - {
+            name: "Windows-MinGW/2019/Static/X64/Release",
+            os: windows-2019,
+            config: Release,
+            cmake_extra_args: -G "CodeBlocks - MinGW Makefiles",
             sudocmd: "",
             artifact_name: "Windows x64",
             cores: 2,
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,7 +51,7 @@ jobs:
             name: "Windows-MinGW/2019/Static/X64/Release",
             os: windows-2019,
             config: Release,
-            cmake_extra_args: -G "CodeBlocks - MinGW Makefiles",
+            cmake_extra_args: "-G \"CodeBlocks - MinGW Makefiles\" -DCMAKE_C_COMPILER=C:/msys64/mingw64/bin/gcc.exe -DCMAKE_CXX_COMPILER=C:/msys64/mingw64/bin/g++.exe -DCMAKE_CXX_FLAGS=\"-lstdc++fs -std=c++17\"",
             sudocmd: "",
             artifact_name: "Windows x64",
             cores: 2,


rath3t:
I think this is not working due to the too old GCC version in this MinGW VM,.

As i found out https://gcc.gnu.org/gcc-9/changes.html  at section Runtime Library (libstdc++).

> Incomplete support for the C++17 Filesystem library and the Filesystem TS on Windows.

This was added in GCC 9. Therefore in earlier version it should not be possible to get this working.

Maybe there is an workaround?

alandefreitas:
@rath3t Thanks for that! It seems like a lot of people have been using MinGW. So this is very helpful.

That's true. GNU 8.1.0 is a little too old. On the other hand, the only thing missing in GNU 8.1.0 is std::filesystem.

Sometimes, the GitHub actions VM includes a more recent version of a compiler but you have to specify the path in `DCMAKE_CXX_COMPILER`. It's a little annoying that this is often not very well documented but I've been able to solve a similar problem like that in the past. If that's the case, there are a few options:

- Looking the at repository with these machines, specifically in the [scripts that install mingw](https://github.com/actions/virtual-environments/blob/24ae664aedf0c3b471057a514208290c26786834/images/win/windows2019.json#L298).
- Using [act](https://github.com/nektos/act) to play with these machines and run these scripts locally. 
- Looking at other CI scripts that use MinGW.

If a more recent version of MinGW is really not available (that's what it looks like), we only have a few options:

- Updating MinGW in the CI script ([this](https://github.com/actions/virtual-environments/blob/24ae664aedf0c3b471057a514208290c26786834/images/win/scripts/Installers/Install-Mingw64.ps1) can serve as a reference) until the VM includes a more recent version. Probably just a matter of changing the chocolatey version in `Choco-Install -PackageName mingw -ArgumentList "--version=8.1.0"`. I imagine they wouldn't even mind if you opened a PR updating MinGW for all machines because they usually try to install the most recent version of all tools. They probably just didn't need to update mingw.
- Using a backup [filesystem](https://github.com/gulrak/filesystem) library when std::filesystem is not available. This is a little more intrusive and might be unnecessary to maintain since std::filesystem is officially in C++17 anyway. On the other hand, this would solve a problem for many other people, even without MinGW.
- Not doing anything until this resolves by itself when they update the VM. The drawback here is we can't ensure matplot++ works with MinGW until they update the VM.

rath3t:
Unfortunalely, it is not possible to test this with `act` since act has no windows support. 
Additionally i tried to update minGW at https://github.com/actions/virtual-environments/pull/3590 which was also not possible due to breaking changes of mingw itself.

I would try to compile it with MSYS an use 
-DCMAKE_C_COMPILER=C:\msys64\mingw64\bin\gcc.exe
-DCMAKE_CXX_COMPILER=C:\msys64\mingw64\bin\g++.exe

I activated the workflows in my fork to not bother your action minutes. I hope that works.

rath3t:
Unfortunately, the test are also run here. I will close this PR and reopen it again when i figured it out

rath3t:
I investigated the problem further: MSYS uses Gcc 10.3.0. Therefore std::filesystem should be available.  

I tried the following:

1. Maybe FindFilesystem.cmake does not play well with Mingw or matplotplusplus?: See   https://github.com/vector-of-bool/CMakeCM/issues/8 
2. Adding -DCMAKE_CXX_FLAGS=\"-lstdc++fs\" does not help
3. Adding -DCMAKE_CXX_FLAGS=\"-enable-libstdcxx-filesystem-ts\" states at configuration "-- Looking for C++ include filesystem - found" but also states later "Cannot run a simple program using std::filesystem" which is quite strange

rath3t:
Now i mimicked the configuration on my own Windows system.
I installed MSYS2 with mingw and i get the following output:

When i compile https://github.com/rath3t/testMingWWorkflow i get the following output on my system:
```
-- The C compiler identification is GNU 10.3.0
-- The CXX compiler identification is GNU 10.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/msys64_2/mingw64/bin/gcc.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/msys64_2/mingw64/bin/g++.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for C++ include filesystem
-- Looking for C++ include filesystem - found
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED - Success
-- Configuring done
-- Generating done
```

in the CI where i copied just the script from https://github.com/alandefreitas/matplotplusplus/blob/master/.github/workflows/build.yml and changed it a little i get:

```
-- The C compiler identification is GNU 10.3.0
-- The CXX compiler identification is GNU 10.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/msys64/mingw64/bin/gcc.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/msys64/mingw64/bin/g++.exe - skipped
D:/a/testMingWWorkflow/testMingWWorkflow
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for C++ include filesystem
-- Looking for C++ include filesystem - found
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED - Failed
-- Performing Test CXX_FILESYSTEM_STDCPPFS_NEEDED
-- Performing Test CXX_FILESYSTEM_STDCPPFS_NEEDED - Failed
-- Performing Test CXX_FILESYSTEM_CPPFS_NEEDED
-- Performing Test CXX_FILESYSTEM_CPPFS_NEEDED - Failed
CMake Error at FindFilesystem.cmake:250 (message):
-- Configuring incomplete, errors occurred!
See also "D:/a/testMingWWorkflow/testMingWWorkflow/build/CMakeFiles/CMakeOutput.log".
  Cannot run a simple program using std::filesystem
See also "D:/a/testMingWWorkflow/testMingWWorkflow/build/CMakeFiles/CMakeError.log".

  You need to update your C++ compiler to use this library
```

I don't have any clue why the CI/VM  behaves here differently.

Any tips?

alandefreitas:
> Unfortunalely, it is not possible to test this with act since act has no windows support

Interesting. It has no MacOS support either. So I guess it's just the Ubuntu VM.

> Unfortunately, the test are also run here. I will close this PR and reopen it again when i figured it out

It's OK to run the tests here. But yes, it's kind of pointless if it's also running in the fork. Thanks.

> Maybe FindFilesystem.cmake does not play well with Mingw or matplotplusplus?

Maybe a more recent version of `FindFilesystem.cmake` would help? I don't know if anything changed since it was included in Matplot++. But it seems to work on your system, so we know doesn't necessarily fail with mingw. 

I don't think there's any reason for that to fail with matplot++, because it's [requesting both](https://github.com/alandefreitas/matplotplusplus/blob/54eed95278ed38dee1aae63ef895756d788597ed/source/matplot/CMakeLists.txt#L1) the Experimental and Final versions of Filesystem. So unless there's a problem with FindFilesystem.cmake, an integration problem with Matplot++ would fail at a later stage, like a missing `#include` or something like that.

> Adding -DCMAKE_CXX_FLAGS="-enable-libstdcxx-filesystem-ts" states at configuration "-- Looking for C++ include filesystem - found" but also states later "Cannot run a simple program using std::filesystem" which is quite strange

That *is* strange. Maybe `FindFilesystem.cmake` is just trying to compile the version with `#include <filesystem>` or something. This seems like a problem with `FindFilesystem.cmake`. This script should already identify we need `-enable-libstdcxx-filesystem-ts` and include this option in the interface target so you don't need `-DCMAKE_CXX_FLAGS="-enable-libstdcxx-filesystem-ts"`. It seems like this is a bug at this point? Maybe `FindFilesystem.cmake` is just not being tested a lot with MinGW.

> CI where i copied just the script from https://github.com/alandefreitas/matplotplusplus/blob/master/.github/workflows/build.yml and changed it a little i get:

Interesting.  I thought `GNU 10.3.0` wouldn't be available.

Does calling/installing `Choco-Install -PackageName mingw -ArgumentList "--version=10.2.0"` help with anything at all?

We shouldn't even need `-enable-libstdcxx-filesystem-ts` in 2021. `std::filesystem` is C++17. 

There might be false negatives with the "`You need to update your C++ compiler to use this library`" message. That was more tolerant in the past, but we just had too many issues where people were just ignoring this message. 

It doesn't seem like a false negative in this case though... because the same script works on your system.

rath3t:
Somehow, I'm not able to reopen this PR since i have forced-pushed stuff.

alandefreitas:
I can't reopen the PR either, because there's another PR from this fork, but I imagine this is the intention.

PR #163, opened 2021-06-18
Add minGW workflow Second Try

rath3t:
This PR should resolve the problems from #161, see also the discussion at: https://github.com/actions/virtual-environments/pull/3590

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
         config:
           # see: https://github.com/actions/virtual-environments
           - {
-            name: "Windows/2019/Static/X86/Release",
+            name: "Windows-MSVC/2019/Static/X86/Release",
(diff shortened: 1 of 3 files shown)


rath3t:
Ok all checks passed but if you look at the Integration test. It says that it is passing. Even if it fails to find the matpolt++ package

alandefreitas:
That looks great. Thanks!

> Ok all checks passed but if you look at the Integration test, it says that it is passing. Even if it fails to find the matplot++ package

The Integration test is allowed to fail for now because finding packages on windows is a little trickier, and we just haven't implemented that yet.

I see it's not generating the artifacts for MinGW yet, but that should be trivial to fix in the future.

rath3t:
Thanks!

alandefreitas:
Looks like the documentation action did its thing now! 😀

PR #184, opened 2021-08-12
Fix "set title" command string when saving image

wjdragon:
- force appending of escaped double quote after font weight detection


alandefreitas:
@wjdragon Thanks and sorry. I thought we had already fixed that for some reason and missed your PR. Is this PR still applicable?

wjdragon:
Yes.  I checked the master branch code and the bug is still there.  Without it, the bold/italicized code will not receive an escaped double quote

PR #195, opened 2021-09-28
Fix axis error in axes_type::ylim()

eagle-dai:
In axes_type::ylim(), x_axis_ should be replaced with y_axis_. Guess this was a copy-paste error.


alandefreitas:
Thank you @eagle-dai !

It's impressive that this bug has been there for so long. Probably because `limits_mode_auto` is usually active on both axes.

Thanks a lot.

PR #213, opened 2021-12-18
Fix warnings on MSVC, changing enum to enum class and initializing va…

nikkaramessinis:
…rs on default constructor


alandefreitas:
@nikkaramessinis Thanks for that! We've been meaning to convert that to some more modern C++ for a long time :D

PR #262, opened 2022-04-25
Add Presets for Installation

codeinred:
This commit adds presets for common use cases, including local and
system installation. It also updates the README to explain the use of
these presets and how they can be used for installation.

If you have CMake 3.21 or greater, you can use the `system` build preset to 
build the package system-wide:

```bash
cmake --preset=system
cmake --build --preset=system
sudo cmake --install build/system
```

Alternatively, if the `CMAKE_PREFIX_PATH` environment variable is set to 
`$HOME/.local`, then you can install it locally. This can be set in `/etc/profile` 
or your shell config. This will not affect discovery of packages installed 
system-wide.

```bash
export CMAKE_PREFIX_PATH="$HOME/.local"
```

This has the advantage of not
requiring sudo, and matplotplusplus will be installed in `$HOME/.local`. 

```bash
cmake --preset=local
cmake --build --preset=local
cmake --install build/local
```

If you're using a version of CMake too old to support presets, then building with
 the system preset is equivilant to:

```bash
cmake -B build/system         \
    -DBUILD_EXAMPLES=OFF      \
    -DBUILD_SHARED_LIBS=ON    \
    -DBUILD_TESTS=OFF         \
    -CMAKE_BUILD_TYPE=Release \
    -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON

cmake --build build/system
```

While building with the local preset is equivilant to:

```bash
cmake -B build/local                      \
    -DBUILD_EXAMPLES=OFF                  \
    -DBUILD_SHARED_LIBS=ON                \
    -DBUILD_TESTS=OFF                     \
    -DCMAKE_BUILD_TYPE=Release            \
    -DCMAKE_INSTALL_PREFIX="$HOME/.local" \
    -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON

cmake --build build/local
```

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,62 @@
+{
+  "version": 3,
+  "cmakeMinimumRequired": {
+    "major": 3,
+    "minor": 19,
+    "patch": 0
+  },
+  "configurePresets": [
+    {
+      "name": "dev",
+      "displayName": "Development Mode",
+      "description": "Build with warnings and debug symbols",
+      "binaryDir": "build",
+      "generator": "Unix Makefiles",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo"
+      }
+    },
+    {
+      "name": "system",
+      "displayName": "Build for system installation",
+      "description": "Bulid for system installation (BUILD_EXAMPLES and BUILD_TESTS are OFF)",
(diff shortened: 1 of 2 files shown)


alandefreitas:
Well... that looks good. Thanks, @codeinred!

PR #263, opened 2022-04-25
Fix spurious toolchain warning

codeinred:
This commit fixes a spurious warning emitted by Matplot++ when building against Matplot++ using `find_package`. This warning is being emitted even when the compiler originally used to build Matplot++ is identical to the one being used to build against the Matplot++ package.

This warning occurs because the Matplot++ Cmake Config file compares the current compiler *version* to the Compiler used to build Matplot++. This always fails, because the version (eg, 11.2.0) is always different from the Compiler ID (In my case, GNU).

The warning was fixed by instead comparing the current Compiler ID to the ID of the compiler used to build Matplot++ (so now if it was built with GNU, and a consumer compiles against it with GNU, there's no warning).

Below is an image of the warning message. Note that the current compiler is also GNU.

image


alandefreitas:
Wow... this is a serious problem.

Thanks again, @codeinred !

codeinred:
Happy to contribute! It’s a good library!

PR #273, opened 2022-06-02
Fix large numbers and add timestamp support

TheGreatRambler:
Fixes using large number for ticks and adds support for rendering timestamps in, for example, xticks

### DIFF
--- a/source/matplot/axes_objects/bars.cpp
+++ b/source/matplot/axes_objects/bars.cpp
@@ -143,6 +143,8 @@ namespace matplot {
 
     std::string bars::data_string() {
         std::stringstream ss;
+        ss.precision(5);
+        ss << std::fixed;
         double m = x_minimum_difference();
         double cw = cluster_width();
         for (size_t bar_group = 0; bar_group < ys_.size(); ++bar_group) {
--- a/source/matplot/backend/gnuplot.cpp
+++ b/source/matplot/backend/gnuplot.cpp
@@ -8,12 +8,12 @@
 #else
 #include <filesystem>
 #endif
+#include <cstdlib>
 #include <iostream>
 #include <matplot/util/common.h>
 #include <matplot/util/popen.h>
 #include <regex>
 #include <thread>
-#include <cstdlib>
 
 #ifdef MATPLOT_HAS_FBUFSIZE
 
@@ -205,9 +205,15 @@ namespace matplot::backend {
             run_command("set terminal " + terminal_ + " position " +
                         num2str(position_[0]) + "," + num2str(position_[1]));
         }
-        if (terminal_has_size_option(terminal_)) {
-            run_command("set terminal " + terminal_ + " size " +
-                        num2str(position_[2]) + "," + num2str(position_[3]));
+        if (terminal_ == "dumb") {
+            run_command("set terminal dumb " + num2str(position_[2]) + " " +
+                        num2str(position_[3]));
+        } else {
+            if (terminal_has_size_option(terminal_)) {
+                run_command("set terminal " + terminal_ + " size " +
+                            num2str(position_[2]) + "," +
+                            num2str(position_[3]));
+            }
         }
     }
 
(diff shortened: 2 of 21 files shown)


TheGreatRambler:
Just a note, what seem like formatting issues are actually the result of your own `.clang-format` file, for example some lines were changed because you have an 80 character column limit and you probably haven't touched these files since you added the column limit

source/matplot/core/axis_type.h
```diff
@@ -106,6 +107,9 @@ namespace matplot {
             return *this;
         }
 
+        bool is_timestamp() const;
```
alandefreitas, line 110: Is the timestamp related to the `ss.precision(5)`?

source/matplot/core/axis_type.h
```diff
@@ -154,6 +158,8 @@ namespace matplot {
         float label_font_size_{11};
         color_array label_color_{0, 0, 0, 0};
 
+        bool is_timestamp_{false};
```
alandefreitas, line 161: Is that all? Do other libraries also work like that?

source/matplot/core/axis_type.cpp
```diff
@@ -247,9 +252,17 @@ namespace matplot {
                 // So if there is no explicit label but there is
                 // a format we want to apply, we also need to explicitly
                 // create a label according to our format.
-                r += "\"" +
-                     escape(num2str(tick_values_[i], tick_label_format_)) +
-                     "\" ";
+                if (is_timestamp_) {
+                    char buff[20] = {0};
+                    time_t now = (time_t)tick_values_[i];
+                    strftime(buff, 20, tick_label_format_.c_str(),
```
alandefreitas, line 258: Will the default `tick_label_format_` work for timestamps?

source/matplot/util/common.h
```diff
@@ -46,6 +46,8 @@ namespace matplot {
 
     template <class T> std::string num2str(Arithmetic<T> num) {
         std::ostringstream ss;
+        ss.precision(5);
```
alandefreitas, line 49: I'm not sure I completely understand the motivation for `precision(5)` at other classes, but it doesn't look like a sensible default for this function. 

For instance, 

```python
str(2.5433653543232154254214321541)
```

gives me 

```python
'2.5433653543232153'
```

in python.

alandefreitas (commented):
Thanks for this! :)

> Fixes using large number for ticks

Why? Were the numbers outside the margins? What is the issue here?

>  and adds support for rendering timestamps

Are these two things related?

> Just a note, what seem like formatting issues are actually the result of your own .clang-format file, for example some lines were changed because you have an 80 character column limit and you probably haven't touched these files since you added the column limit

That's fine. We don't reformat the whole thing whenever parameters change. We only reformat files as we change them.

source/matplot/axes_objects/bars.cpp
```diff
@@ -143,6 +143,8 @@ namespace matplot {
 
     std::string bars::data_string() {
         std::stringstream ss;
+        ss.precision(5);
```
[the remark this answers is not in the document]
TheGreatRambler: The default `stringstream` precision prevented rendering of very small numbers

source/matplot/backend/gnuplot.cpp
```diff
@@ -205,9 +205,15 @@ namespace matplot::backend {
             run_command("set terminal " + terminal_ + " position " +
                         num2str(position_[0]) + "," + num2str(position_[1]));
         }
-        if (terminal_has_size_option(terminal_)) {
-            run_command("set terminal " + terminal_ + " size " +
-                        num2str(position_[2]) + "," + num2str(position_[3]));
+        if (terminal_ == "dumb") {
+            run_command("set terminal dumb " + num2str(position_[2]) + " " +
+                        num2str(position_[3]));
+        } else {
```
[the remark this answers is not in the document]
TheGreatRambler: Okay wait this shouldn't be here lmao, but I do remember why I put this here, the dumb terminal is super funny and actually has a nonstandard way of setting the size, as you can see here, but it's not relevant to this PR

source/matplot/core/axis_type.h
TheGreatRambler: `timestamp(true)` enables timestamp rendering, so `ax->xtickformat ("%m/%d/%Y %H:%M:%S");` can be used for example. `ss.precision()` is relevant to large numbers

source/matplot/axes_objects/bars.cpp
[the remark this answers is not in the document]
TheGreatRambler: Also I believe (these are some old changes I wanted to PR so it's been a while) gnuplot would start using scientific notation without that change

source/matplot/core/axis_type.cpp
TheGreatRambler: No, because `strftime` supports formatting the hour, minute, second, etc using a millisecond input. `num2str` does not do this

TheGreatRambler:
Okay so the dumb terminal should be removed entirely, and `ss.precision` is probably not required (though `ss << std::fixed` is absolutely required to support large numbers)

TheGreatRambler:
I should mention this explicitely as well, the issue I was referring to was large numbers being converted to scientific notation by default, which is somewhat annoying especially in the y axis

TheGreatRambler:
Glad it was useful! Got a bit preoccupied on other code, because of the font issue I shelved this project

PR #281, opened 2022-07-30
Fix incorrect min/max values.

csernib:


alandefreitas:
Great. Thanks for identifying and fixing this.

I think these values are not effectively used that often so we never noticed this.

PR #288, opened 2022-08-07
Fix to use localtime for Linux

kwoodle:
Changed axis_type.cpp to use localtime since localtime_s doesn't work in Linux. Added     add_definitions(-D_CRT_SECURE_NO_WARNINGS) to source/matplot/CMakeLists.txt to make localhost work on Windows. I usually work in Linux. Not sure if this breaks anything on Windows, although I did build ContinuousBuild target and ran some examples.


source/matplot/CMakeLists.txt
```diff
@@ -159,6 +159,9 @@ endif ()
 #######################################################
 ### Library options                                 ###
 #######################################################
+if(MSVC)
+    add_definitions(-D_CRT_SECURE_NO_WARNINGS)
```
alandefreitas, line 163: target_compile_definitions?

source/matplot/core/axis_type.cpp
```diff
@@ -256,9 +256,9 @@ namespace matplot {
                 if (is_timestamp_) {
                     char buff[20] = {0};
                     time_t now = (time_t)tick_values_[i];
-                    struct tm buf;
-                    localtime_s(&buf, &now);
-                    strftime(buff, 20, tick_label_format_.c_str(), &buf);
+                    struct tm* buf;
```
alandefreitas, line 259: Isn't `#ifdef WIN32 ...` simpler than the external `-D_CRT_SECURE_NO_WARNINGS` here?

alandefreitas (commented):
Great. We could squash this into a single commit.

source/matplot/core/axis_type.cpp
kwoodle: I've added macros and thread-safe localtime_r to axis_type.cpp, removed the change to CMakeLists.txt, and pushed the changes to my fork. Do I need to do anything else to integrate the changes? Also, when I run the .exe examples on Windows, gnuplot has trouble reading input. Hope you can check this. Thanks for your suggestions.

source/matplot/core/axis_type.cpp
```diff
@@ -257,7 +257,11 @@ namespace matplot {
                     char buff[20] = {0};
                     time_t now = (time_t)tick_values_[i];
                     struct tm buf;
+#if defined(__unix__)
+                    localtime_r(&now, &buf);
+#elif defined(_WIN32)
                     localtime_s(&buf, &now);
+#endif
```
alandefreitas, line 264: Nice. We need a third alternative when both conditions are false.

```cpp
#if defined(__unix__)
                    localtime_r(&now, &buf);
#elif defined(_WIN32)
                    localtime_s(&buf, &now);
#else
                    // use std::localtime from https://en.cppreference.com/w/cpp/chrono/c/localtime
#endif
```

We also need an equivalent pattern to include the proper headers:

```cpp
#if defined(__unix__)
#include <header with localtime_r>
#elif defined(_WIN32)
#include <header with localtime_s>
#else
#include <ctime>
#endif
```

source/matplot/CMakeLists.txt
```diff
@@ -159,7 +159,6 @@ endif ()
 #######################################################
 ### Library options                                 ###
 #######################################################
-
```
alandefreitas, line 162: Can't you just rollback this non-change?

source/matplot/core/axis_type.cpp
alandefreitas: > I've added macros and thread-safe localtime_r to axis_type.cpp, removed the change to CMakeLists.txt, and pushed the changes to my fork. 

Nice. Could you squash these commits?

> Do I need to do anything else to integrate the changes? 

I think we are missing the `std::localtime` default now.

> Also, when I run the .exe examples on Windows, gnuplot has trouble reading input. Hope you can check this. 

Sure. That's unrelated but I'll have a look.

Thanks for the PR.

source/matplot/core/axis_type.cpp
```diff
@@ -257,7 +263,13 @@ namespace matplot {
                     char buff[20] = {0};
                     time_t now = (time_t)tick_values_[i];
                     struct tm buf;
+#if defined(__unix__)
+                    localtime_r(&now, &buf);
+#elif defined(_WIN32)
                     localtime_s(&buf, &now);
+#else
+                    buf = std::localtime(&now);
```
alandefreitas, line 271: We have a bug here:

https://github.com/alandefreitas/matplotplusplus/runs/7770952743?check_suite_focus=true#step:5:20

alandefreitas:
Nice! Thanks!

kwoodle:
You're too kind!

PR #303, opened 2022-11-01
macos install gnuplot --don't run as root

saxbophone:
Homebrew isn't designed to be run as root:

image


alandefreitas:
Yes. That makes sense. Thanks!

PR #323, opened 2023-01-30
Correct cmake variables

david-x64:
I noticed that cmake `install` target was not working correctly.
So I investigated and found that `BUILD_INSTALLER` variable was recently renamed but not updated.

After more investigation, I found other variables that also were not renamed completely.

This pull request will correct cmake variables by completing the rename for:
 - `BUILD_INSTALLER`
 - `BUILD_FOR_DOCUMENTATION_IMAGES`
 - `BUILD_HIGH_RESOLUTION_WORLD_MAP`
 - `BUILD_WITH_PEDANTIC_WARNINGS`
 - `BUILD_EXPERIMENTAL_OPENGL_BACKEND`

### DIFF
--- a/cmake/functions/target_options.cmake
+++ b/cmake/functions/target_options.cmake
@@ -11,7 +11,7 @@ endfunction()
 
 # @brief Maybe enable pedantic warnings for a target
 function(maybe_target_pedantic_warnings TARGET_NAME)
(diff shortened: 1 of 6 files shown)


alandefreitas:
Oh! Thanks!

PR #337, opened 2023-05-03
fix: install matplot_opengl in CMake

aminya:
Fixes the missing matplot_opengl install

### DIFF
--- a/source/matplot/CMakeLists.txt
+++ b/source/matplot/CMakeLists.txt
@@ -90,6 +90,8 @@ add_library(matplot
         freestanding/plot.h
 )
 
+set(TARGETS matplot)
+
 # Target aliases
 add_library(Matplot++::matplot ALIAS matplot)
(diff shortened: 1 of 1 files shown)


alandefreitas:
This target is very experimental and missing almost all features. I didn't think anyone wanted that installed.

In any case, I believe it would still need to find the dependencies in the cmake config file to work properly.

aminya:
The installation only happens when the experimental option is enabled. Otherwise, it will not be installed

alandefreitas:
> The installation only happens when the experimental option is enabled. Otherwise, it will not be installed

Yes. That's fine.

Have you tried using what it installs locally? Did it work? I would expect a few `find_dependency` in the config template for that to work.

aminya:
Yes, I tested it. It actually works with vcpkg as well the platforms they test:
https://github.com/microsoft/vcpkg/pull/31229

alandefreitas:
Thanks! :)

PR #352, opened 2023-06-07
feat: support OpenGL dependencies with config

aminya:
This adds `find_package` for `glad` and `glfw3` using `CONFIG`. This allows the package to be used better with `vcpkg`.


source/matplot/CMakeLists.txt
```diff
@@ -214,40 +214,49 @@ if (MATPLOTPP_BUILD_EXPERIMENTAL_OPENGL_BACKEND)
...
-            target_link_libraries(glad INTERFACE ${GLAD_LIBRARIES})
-        endif ()
-    endif ()
+    find_package(glad CONFIG)
+    if (NOT glad_FOUND)
+        find_package(GLAD QUIET)
+        if (NOT GLAD_FOUND AND NOT TARGET glad)
+            # Use CPM only if not found, to avoid ODR violations
+            # find_package(GLAD REQUIRE) would suffice if it worked well
+            FetchContent_Declare(glad GIT_REPOSITORY https://github.com/Dav1dde/glad.git GIT_TAG df8e9e16110b305479a875399cee13daa0ccadd9)
```
alandefreitas, line 226: You can update the GIT_TAG to match your vcpkg version if you want.

source/matplot/CMakeLists.txt
```diff
@@ -214,40 +214,49 @@ if (MATPLOTPP_BUILD_EXPERIMENTAL_OPENGL_BACKEND)
...
-    find_package(glfw3 QUIET)
-    if (NOT GLFW3_FOUND AND NOT TARGET glfw)
-        # Use CPM only if not found, to avoid ODR violations
-        # find_package(glfw3 REQUIRE) would suffice if it worked well
-        FetchContent_Declare(glfw3 GIT_REPOSITORY https://github.com/glfw/glfw.git GIT_TAG 3.3.2)
-        FetchContent_MakeAvailable(glfw3)
-    endif ()
+    find_package(glfw3 CONFIG)
+    if (NOT glfw3_FOUND)
+        find_package(glfw3 QUIET)
```
alandefreitas, line 244: This is repeating the same package from two lines above. I don't think it's ever GLFW3.

aminya: The search procedure is different though (without config). However, I think it doesn't make a difference.

source/matplot/CMakeLists.txt
```diff
@@ -214,40 +214,49 @@ if (MATPLOTPP_BUILD_EXPERIMENTAL_OPENGL_BACKEND)
...
-        # FindGLAD does not usually create a target, so we create an interface target
-        if (NOT TARGET glad)
-            add_library(glad INTERFACE)
-            target_include_directories(glad INTERFACE ${GLAD_INCLUDE_PATH})
-            target_link_libraries(glad INTERFACE ${GLAD_LIBRARIES})
-        endif ()
-    endif ()
+    find_package(glad CONFIG)
+    if (NOT glad_FOUND)
+        find_package(GLAD QUIET)
```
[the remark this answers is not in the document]
aminya: I usually review the diffs by hiding whitespace changes. I'll address this just in case.

source/matplot/CMakeLists.txt
alandefreitas: Oh... there are config and module versions. In that case, I think the module version falls back to config mode, but I don't know if config mode falls back to module mode. I don't think so, because they are discontinuing module mode.

alandefreitas: Just checked the docs. Yes. Just removing config from the first find_package call is good enough. It will fallback to config if it can't find it and the user can even choose the revert this order.

https://cmake.org/cmake/help/latest/command/find_package.html#id6

aminya:
@alandefreitas Addressed all the comments.

alandefreitas:
That looks good. Thanks!

PR #356, opened 2023-06-17
Fix rticks(), rtickformat(), rticklabels() and rtickangle() being ignored

j0hnnybash:
The `rtick*` group of functions previously had no effect unless the the r-axis was manually made visible. The r-axis is invisible by default and instead the major grid is enabled by default (which is much more helpful for polar plots), however this had the undesirable fact that  code block that handles the `*tick*()` functions for arbitrary axis would not be executed.

Another related issue which is also fixed by this change is that both `ax->rticklabels({});`  and `ax->rticks({});` would not hide the tick labels, since hiding tick labels without also hiding the grid requires special handling.

### DIFF
--- a/source/matplot/core/axes_type.cpp
+++ b/source/matplot/core/axes_type.cpp
@@ -429,7 +429,15 @@ namespace matplot {
                                         std::string axis_name,
                                         bool minor_ticks = false) {
             // cb is the only axis we don't unset if tics are empty
-            if (ax.visible() && !ax.tick_values().empty()) {
+            // r-axis labels should still be handled even if axis is invisible since we use the grid
+            if ((ax.visible() || axis_name == "r") && !ax.tick_values().empty()) {
+                if (axis_name == "r" && !ax.visible()) {
+                    // Hide the r-axis but not the grid.
+                    // We can't completely unset the tics or the grid does not
+                    // appear properly
+                    run_command("unset " + axis_name + "axis");
+                    run_command("set " + axis_name + "tics scale 0");
+                }
                 if (ax.geographic()) {
                     run_command("set " + axis_name + "tics geographic");
                 }
@@ -461,13 +469,21 @@ namespace matplot {
                         run_command("set m" + axis_name + "tics 2");
                     }
                 }
-            } else {
+            } else { // (!visible && !r) || empty
                 if (axis_name == "r") {
                     // Hiding the r-axis works differently.
                     // We can't completely unset the tics or the grid does not
                     // appear properly
                     run_command("unset " + axis_name + "axis");
                     run_command("set " + axis_name + "tics scale 0");
+                    if (ax.tick_values().empty()) {
+                        // Note: Disabling tick labels without also
+                        // disabling the grid works via 'set format'
+                        // as per gnuplot manual "xticks" (see tic
+                        // labels). using 'set xticks () ...' would be
+                        // insufficient.
+                        run_command("set format " + axis_name + " \"\"");
+                    }
                 } else if (axis_name == "cb") {
                     // the colorbar / colorbox has a special command to unset it
                     // if only ticks are empty we still want to show the box


alandefreitas:
> undesirable fact that code block that handles the *tick*() functions for arbitrary axis would not be executed

I'm not sure I understand this sentence. Would you have an example?

Obs: would you mind using conventional commits and present tense? That allows commits to be properly incorporated into the changelog. Ex: `fix: r-ticks are not ignored when ...`

j0hnnybash:
> I'm not sure I understand this sentence. Would you have an example?

Reading my own comment back, I realize it's neither understandable not grammatically correct...

I'll try to come up with some examples (also for the documentation) based on my test code.

examples/polar_plots/polarplot/polarplot_10.cpp
```diff
@@ -0,0 +1,33 @@
...
+    const auto uniform_density = 1./(2.0 * pi);
+
+    auto fig = figure(true);
+    auto ax = fig->add_subplot(1, 1, 0);
+
+    ax->rticklabels({"uniform", "max (κ=2.)"});
+    ax->rticks({uniform_density, max});
+    ax->rtickangle(60.);
+    ax->polarplot(rhos);
+
```
j0hnnybash, line 30: Without the fix, it is necessary to add `ax->r_axis().visible(true);` _after_ the `polarplot()` call in order for the custom tick labels to show up.

alandefreitas:
Oh... you included complete examples. :) I meant just small inline examples of what you mean in the PR. The new examples are fine, but they require new images for the docs, etc... But I think I get the point now.

source/matplot/core/axes_type.cpp
```diff
@@ -429,7 +429,15 @@ namespace matplot {
                                         std::string axis_name,
                                         bool minor_ticks = false) {
             // cb is the only axis we don't unset if tics are empty
-            if (ax.visible() && !ax.tick_values().empty()) {
+            // r-axis labels should still be handled even if axis is invisible since we use the grid
```
alandefreitas, line 432: Wouldn't it be more semantically correct if setting up the grid made it visible? Because this way the thing is as if both visible and invisible. The user will be literally seeing the thing and `visible()` would give him `false`.

j0hnnybash: Well the `visible()` correctly refers to the axis itself (which is hidden) and not the grid, however for polar plots the library basically treats the grid as the axis and disables the axis by default (as just a single r-axis is rather unhelpful for a circular plot). 
Semantically it could make sense to hide gnuplot's notion of the r-axis from the user entirely and treat gnuplot's polar grid as the r-axis instead, but that sounds like a major refactor...


> The user will be literally seeing the thing and `visible()` would give him `false`.

This is already the case: `polarplot()` sets `r_axis.visible(false)` because it refers to the axis, and not to the visible grid which however easily might be mistaken for the r_axis...

examples/polar_plots/polarplot/polarplot_10.cpp
j0hnnybash: Note: the resulting images are not identical: `ax->r_axis().visible(true);` has the additional (obvious) effect of making the r-axis visible which, at least in my case, might be undesirable.

source/matplot/core/axes_type.cpp
j0hnnybash: Maybe this illustration clarifies things:

Output of `example_polarplot_10` with the fix applied:
polar_plot_rticks_good_img

Output of `example_polarplot_10` with the addition of `ax->r_axis().visible(true);`, result independent of fix:
polar_plot_rticks_bad_img


P.S.: The r-tick label overlaps the t-tick degree label presumably because of my broken HiDPI screen fractional scaling setup (fonts only scaling...). When running the example on my non-HiDPI laptop screen the labels do not overlap.

alandefreitas:
> Matplotplusplus / MacOSX/10.15/Static/X64/Release (pull_request) Failing after 5m

Uh-oh

alandefreitas:
> The OSX build seems to have trouble finding [std::cyl_bessel_i](https://en.cppreference.com/w/cpp/numeric/special_functions/cyl_bessel_i), but this should be available in <cmath> with C++17.

Yes. It's common for some STL features never to get really implemented depending on the demand for them. The C++ garbage collector got approved and deprecated without ever being really implemented by anyone.

> Maybe I should dumb down the example...

You can remove the examples if you want. We just needed them for the discussion here. If we're keeping these examples, we need to add them to README.md so they can be picked up by the documentation generator.

alandefreitas:
Could you please rebase this PR on top of master or develop? The PR is now behind these branches because there has been a release.

j0hnnybash:
FYI: I rebased onto master.

alandefreitas:
> FYI: I rebased onto master.

Great! Thanks!

PR #358, opened 2023-06-17
Fix polarhistograms with many bins

j0hnnybash:
This PR consists of two fixes which both affected the correct display of polarhistograms with more than 89 bins.
Note: The second commit also increases the minimum polarhistogram resolution from 89 to 359 line segments.

#### Fix 1: Fix `linspace(a,b,1)` resulting in `NaN` element

This underlying issue caused polarhistograms with more than `89` bins to
not be displayed at all.

Previously passing `n=1` to `linspace` resulted in a one element vector
containing `NaN`. With this change it will result in a one element
vector containing the first ("lower") bound.  This is the natural
solution given the current implementation and matches the behavior of
`numpy.linspace`, unfortunately however it does not match the behavior
of MATLAB `linspace` which returns the "upper" bound instead.

#### Fix 2: Ensure at least two points (one circle segment) is used per polarhistogram petal

Change petal resolution to the larger one of (a) `360` points (`359`
segments) per circle or (b) `n_histogram_bins+1`
points (`n_histogram_bins` segments). Ensuring at least one segment per
bin petal.

Previously polarhistograms had a fixed resolution of `90` points (`89`
segments) per circle. Specifying more than `89` bins would result in
either an empty plot (due to a bug in `linspace`, see Fix 1) or too narrow
petals (with the `linspace` bug fixed).

### DIFF
--- a/source/matplot/axes_objects/histogram.cpp
+++ b/source/matplot/axes_objects/histogram.cpp
@@ -102,9 +102,10 @@ namespace matplot {
 
         } else {
             // resolutions of the petals
-            constexpr double points_per_circle = 90;
+            constexpr double points_per_circle = 360;
             const double n_histogram_bins = bin_edges_.size() - 1.;
-            const double points_per_bin = points_per_circle / n_histogram_bins;
+            // ensure at least one line segment (2 points) per petal.
+            const size_t points_per_bin = static_cast<size_t>(max(2.0, points_per_circle / n_histogram_bins));
             if (!stairs_only_) {
                 for (size_t i = 0; i < values_.size(); ++i) {
                     // make a petal for each value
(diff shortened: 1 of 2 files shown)


alandefreitas:
> This is the natural solution given the current implementation and matches the behavior of numpy.linspace, unfortunately however it does not match the behavior of MATLAB linspace which returns the "upper" bound instead.

Yes. That's reasonable. Even though matlab was the reference for both implementations. In C++, both functions are kind of silly and should actually be some polyfill for std::ranges::views::iota or some variantion of that but we don't have that right now anyway.

> Change petal resolution to the larger one of (a) 360 points (359 segments) per circle or (b) n_histogram_bins+1 points (n_histogram_bins segments).

Yes. These precision changes are OK to apply. The proper solution would be dynamic precision based on figure properties but gnuplot (gnuplot again) through pipes doesn't give us that level of control. Until some SDL backend solution comes up, adjusting precision on user needs is the best we can do and it's very reasonable.

j0hnnybash:
I updated the commit messages to match the projects convention (as requested  [here](https://github.com/alandefreitas/matplotplusplus/pull/356#issuecomment-1595869567)).

Issue #370, opened 2023-09-15
"pngcairo" doesn't use "size" option


ino-josh:
**Feature category**
- [ ] *enhancement - build system*
- [x] *enhancement - backends*
- [ ] *enhancement - build system*
- [ ] *enhancement - documentation*
- [ ] *enhancement - plot categories*

**The problem**
When saving, if the format is pngcairo, the `size` option is not passed to gnuplot.
So when you do `save("/test.png")` or `save("/test", "pngcairo")`, the resulting image is always 640x480.

**The solution I'd like**
In `gnuplot.cpp`:
```cpp
    bool gnuplot::terminal_has_size_option(const std::string &t) {
        // Terminals that have the size option *in the way we expect it to work*
        // This includes only the size option with {width, height} and not
        // the size option for cropping or scaling
        SV_CONSTEXPR std::string_view whitelist[] = {
            "qt",      "aqua",     "caca",    "canvas", "eepic",
            "emf",     "gif",      "jpeg",    "pbm",    "png",
            "sixelgd", "tkcanvas", "windows", "wxt",    "svg"};
        return std::find(std::begin(whitelist), std::end(whitelist), t) !=
               std::end(whitelist);
    }
```
simplty add `"pngcairo"` to the list.

According to gnuplot documenation:
```
set term pngcairo
                     {{no}enhanced} {mono|color}
                     {{no}transparent} {{no}crop} {background <rgbcolor>
                     {font <font>} {fontscale <scale>}
                     {linewidth <lw>} {rounded|butt|square} {dashlength <dl>}
                     {pointscale <ps>}
                     {size <XX>{unit},<YY>{unit}}
```
and
```
By default the X and Y sizes are taken to be in pixels...
```
I believe that satisfies the statement `Terminals that have the size option *in the way we expect it to work*`

**Alternatives I've considered**
Using the "png" terminal, but it looks like trash compared to "pngcairo" :)

**Additional context**
I am submitting this as an issue as opposed to a pull request in case there is some reason this is the way that it is intentionally, and I just don't know that reason.

alandefreitas:
That makes sense.

> I am submitting this as an issue as opposed to a pull request in case there is some reason this is the way that it is intentionally, and I just don't know that reason.

No. It's not really intentional.

PR #371, opened 2023-09-18
fix: "pngcairo" should use size option

ino-josh:
"pngcairo" missing from list is not intentional: #370


alandefreitas:
Thanks :)

Issue #376, opened 2023-10-23
Quiet mode does not work on windows when matplot is called from a GUI application


geral-victor:
**Bug category**

- [ ] *bug - compilation error*
- [ ] *bug - compilation warning*
- [ ] *bug - runtime error*
- [ ] *bug - runtime warning*
- [x] *bug - logic error*

**Describe the bug**
When creating a figure from a GUI app on windows, in order to generate ***quietly*** a plot and save the resulting output as an image, a terminal gets opened.

The problem does not exists if we run the library in a console program.

**Steps to Reproduce**
```C++
{
   auto f = matplot::figure(true); //opens a terminal that is visible, despite having chosen quiet mode.
   ...
   f->save(filename);
}
```

**Dirty fix**
```C++
{
   auto f = matplot::figure(true); //opens a terminal that is visible, despite having chosen quiet mode.
   ...
   f->save(filename);
   f->backend({}); // dirty patch to close the console opened when matplot++ is called from a GUI, and not a console
}
```

**Platform**

- [ ] *cross-platform issue - linux*
- [x] *cross-platform issue - windows*
- [ ] *cross-platform issue - macos*

**Environment Details:**
 - OS: Windows 10

Is seems that the issue is related to the use of `_popen` : https://github.com/alandefreitas/matplotplusplus/blob/a40344efa9dc5ea0c312e6e9ef4eb7238d98dc12/source/matplot/backend/gnuplot.cpp#L76

It seems that `_popen` cannot hide windows, as we can read here https://stackoverflow.com/a/10020726/18371648

So we should use `CreateProcess` instead of `_popen`

PR #377, opened 2023-10-24
Prevent console window to open in GUI applications on Windows

geral-victor:
Fix for https://github.com/alandefreitas/matplotplusplus/issues/376

On Windows, in a GUI app using this library, a console window gets opened when using `_popen`.
It seems (SO reference [here](https://stackoverflow.com/questions/10020653/is-there-any-way-of-stopping-popen-opening-a-dos-window/10020726#10020726)) that the only way to avoid this is to use `CreateProcess`.

The use case that this PR fixes is when you want to use the lib in a GUI app on Windows, to generate figures that you save into files. I think that is what is called non-interactive mode.

### DIFF
--- a/source/matplot/util/popen.h
+++ b/source/matplot/util/popen.h
@@ -6,18 +6,19 @@
 #define MATPLOTPLUSPLUS_POPEN_H
 
 #ifdef _WIN32
-#include <io.h>
-#include <stdio.h>
+#include <stdio.h> //needed for FILE type
 #include <windows.h>
 
-// in a GUI application, _popen pops up a window. This version does not.
-FILE *hiddenPopen(const char *command, const char *mode);
+namespace matplot::detail {
+    // in a GUI application, _popen pops up a window. This version does not.
+    FILE *hiddenPopen(const char *command, const char *mode);
 
-// close the handle opened by hiddenPopen
-int hiddenPclose(FILE *file);
+    // close the handle opened by hiddenPopen
+    int hiddenPclose(FILE *file);
+} // namespace matplot::detail
 
-#define PCLOSE hiddenPclose
-#define POPEN hiddenPopen
+#define PCLOSE ::matplot::detail::hiddenPclose
+#define POPEN ::matplot::detail::hiddenPopen
 #define FILENO _fileno
 #else
 #define PCLOSE pclose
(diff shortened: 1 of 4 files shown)


source/matplot/util/popen.h
```diff
@@ -6,9 +6,18 @@
 #define MATPLOTPLUSPLUS_POPEN_H
 
 #ifdef _WIN32
+#include <io.h>
```
alandefreitas, line 9: Do we need these in the header file? Can't these just go to the cpp file?

alandefreitas (commented):
This makes sense. Thanks.

The `hiddenPopen` and `hiddenPclose` functions need to a namespace thought. It can be `matplot::detail` and then `#define PCLOSE ::matplot::detail::hiddenPclose` and so on.

source/matplot/util/popen.h
geral-victor: We can move `#include <io.h>` in the cpp file. However, `#include <stdio.h>` must be included in the header. I'm updating the PR following your comments.

geral-victor:
I am transferring the functions to `matplot::detail`, and I will update the PR soon following your suggestion. In regards to the interactive mode, it seems to work fine on my end with this modification, but it's not the main mode I use.

alandefreitas:
Nice

source/matplot/util/popen.h
alandefreitas: Does it work if we forward-declare `FILE` in the header file, then use `FILE*` in the header file, then include `stdio.h` in the cpp file?

geral-victor: I woulnd't to this, to avoid adding a lot of `reinterpret_cast` in the implementation (because of the namespace). What do you think?

alandefreitas: Maybe I'm not miscommunicating. There's no `reinterpret_cast` in what I'm proposing.

popen.cpp:

```
#include <io.h>
```

becomes:

```
#include <io.h>
#include <stdio.h>
```

popen.h:

```
#include <stdio.h> //needed for FILE type
#include <windows.h>
```

becomes

```
#include <windows.h>
struct FILE;
```

but I just noticed this is incorrect for some other reason: The standard just states that FILE is 

> an object type capable of recording all the information needed to control a stream"; it's up to the implementation whether this is a typedef of a struct (whose name you don't know anyway), or something else.

In any case, it doesn't make much sense to avoid `stdio.h` when `windows.h`, which is a gazillion times heavier, is right next it.

Issue #393, opened 2024-02-20
Add SOVERSION to shared library


badshah400:
**Feature category**
- [x] *enhancement - build system*
- [ ] *enhancement - backends*
- [ ] *enhancement - documentation*
- [ ] *enhancement - plot categories*

**The problem**
This is a request to add an `SOVERSION` to the shared library `libmatplot` that may only be incremented to indicate breakage or interface changes. According to [TLDP](https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html),

> Every shared library has a special name called the "soname''. The soname has the prefix "lib'', the name of the library, the phrase ".so'', followed by a period and a version number that is incremented whenever the interface changes

**The solution I'd like**

Ideally the soversion will be the major version, i.e. `1` for `matplot++` 1.2.x. This patch does it:

```diff
Index: matplotplusplus-1.2.1/source/matplot/CMakeLists.txt
===================================================================
--- matplotplusplus-1.2.1.orig/source/matplot/CMakeLists.txt
+++ matplotplusplus-1.2.1/source/matplot/CMakeLists.txt
@@ -98,7 +98,7 @@ target_include_directories(matplot
     PUBLIC $<BUILD_INTERFACE:${MATPLOT_ROOT_DIR}/source>
            $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 
-set_target_properties(matplot PROPERTIES VERSION ${PROJECT_VERSION})
+set_target_properties(matplot PROPERTIES VERSION ${PROJECT_VERSION} SOVERSION ${PROJECT_VERSION_MAJOR})
 
 # Shared library symbol export
 include(GenerateExportHeader)

```

**Alternatives I've considered**

NA

**Additional context**

I am a packager for openSUSE and am working on submitting matplot++ packages to the distro's official repository.

Thanks a lot for your very useful library.

alandefreitas:
Thanks for the issue. Yes. This makes sense.

PR #394, opened 2024-02-21
Add SOVERSION to shared library

badshah400:
Set the soversion of the matplot shared library to the major version number of the project. This provides a guarantee that breaking changes to the library interface shall only be made when the major version is incremented.

Fixes issue #393.


alandefreitas:
That's great. Thanks!

Issue #427, opened 2024-10-17
std::system_error on Windows


bitsbakery:
**Bug category**

- [*] *bug - runtime error*

**Describe the bug**
Cannot launch any example, I get the following std::system_error:
```
struct pipe_deleter {
        int operator()(FILE* pipe) const {
            if (int status = PCLOSE(pipe); status != -1)
                return status;
            //throw std::system_error{errno, std::system_category(), "pclose"};
        }
    };
```
After commenting out "throw std::system_error", the examples start working. 

**Steps to Reproduce**
1. Build [matplotplusplus](https://github.com/alandefreitas/matplotplusplus) in Development Mode
2. Install gnuplot v6
3. Try to launch any example

**Platform**

- [*] *cross-platform issue - windows*

alandefreitas:
I don't know. Maybe a different status code is not that bad.

@mikucionisaau ?

mikucionisaau:
Oh well, non-POSIX OS does not behave like POSIX says :-(

I see that there is a nice `popen` and `pclose` wrapper in `popen.cpp` for `_WIN32`, but it is not setting the `errno` like POSIX `pclose` does, so the error gets lost (I get "Success" with mingw64 and wine setup on Linux).

I do not see any other signs of cross-compilation attempts in this repo, so I was wondering how do you compile? What OS/toolchain/compiler do you use?

bitsbakery:
> What OS/toolchain/compiler do you use?

No cross-compilation, I compile natively on Windows 10, using 
```
Microsoft (R) C/C++ Optimizing Compiler Version 19.39.33523 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.
```
Adjusted CMAKE_CXX_COMPILER in cmake presets to "cl" (this is MSVC).

mikucionisaau:
Good news: I found [_popen](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/popen-wpopen?view=msvc-170) in Windows, which works, so `popen.cpp` is overly complicated and can be retired, unless there is some old Windows that needs it.
I'll make a PR for this.

mikucionisaau:
Bad news: popen.cpp was created to hide console window to fix https://github.com/alandefreitas/matplotplusplus/issues/376

alandefreitas:
popen is always causing minor issues over the years 🤣

mikucionisaau:
[...]

PR #428, opened 2024-10-18
Fix error reporting of popen and pclose for Windows (issue 427)

mikucionisaau:
Attempt to fix [issue 427](https://github.com/alandefreitas/matplotplusplus/issues/427)

### DIFF
--- a/source/matplot/backend/gnuplot.cpp
+++ b/source/matplot/backend/gnuplot.cpp
@@ -80,8 +80,8 @@ namespace matplot::backend {
 
         // Check if everything is OK
         if (perr != 0 || !pipe_.opened()) {
-            std::cerr << "Opening the gnuplot pipe_ failed: " 
-                << std::strerror(perr) << std::endl;
+            std::cerr << "Opening the gnuplot pipe_ failed:\n";
+            std::cerr << pipe_.error() << std::endl;
             std::cerr << "Please install gnuplot 5.2.6+: http://www.gnuplot.info"
                 << std::endl;
         }
--- a/source/matplot/backend/gnuplot.cpp
+++ b/source/matplot/backend/gnuplot.cpp
@@ -72,16 +72,16 @@ namespace matplot::backend {
         // Open the gnuplot pipe_
         int perr;
         if constexpr (windows_should_persist_by_default) {
-            perr = popen2("gnuplot --persist", "w", &pipe_);
+            perr = pipe_open(&pipe_, "gnuplot --persist", "w");
         } else {
-            perr = popen2("gnuplot", "w", &pipe_);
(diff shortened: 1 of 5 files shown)


alandefreitas:
Oh... That's interesting.

mikucionisaau:
@alandefreitas I think the core issue is that `popen` is hackish: 
1) OS does a lot of strange things behind the scenes, like recovering the child process handle from `FILE` handle.
2) the previous implementation used some weird (internal?) features I could not figure out how they work. Quite possible that something has changed in the C library, and thus recovering the handle to the child process does not work anymore.

So instead, I rewrote `popen` functionality from scratch (it appears that it is quite common to replace `popen` with custom implementation, even for Linux) and wrapped `FILE` and process handle into C++ classes.

On one hand, one can make C API with a custom data structure (looks ugly, see commit history).
On the other hand, one can even wrap those pipe classes into `std::istream` and `std::ostream`, expose the rich API and hide the process pipe complexity, but the gnuplot backend wrapper does not need that much.

It would be nice to test this PR on a Mac before merging.

source/matplot/backend/gnuplot.cpp
```diff
@@ -323,11 +321,9 @@ namespace matplot::backend {
         static std::string terminal_type;
         const bool dont_know_term_type = terminal_type.empty();
         if (dont_know_term_type) {
-            terminal_type =
-                run_and_get_output("gnuplot -e \"show terminal\" 2>&1");
-            terminal_type = std::regex_replace(
-                terminal_type, std::regex("[^]*terminal type is ([^ ]+)[^]*"),
-                "$1");
```
mikucionisaau, line 330: @alandefreitas `std::regex` is ridiculously slow, consider `string_view`:
https://quick-bench.com/q/IUXYt_GIlS3VW8Gh4hJg8HLVQj4

alandefreitas: Yes. A custom parser would be better here, also in terms of compilation time.

alandefreitas:
Oh... Thanks! This ended up being quite complex. Nice work. Is it good to go? It looks good to me.

CI would check Apple Clang, but it didn't run for some reason. It seems it needs some maintenance anyway because the latest runs broke in https://github.com/alandefreitas/matplotplusplus/actions. That's expected from time to time. It shouldn't block this PR because it's unrelated to it, though.

mikucionisaau:
I just fixed error reporting for MSVC, which is very weird.
Also tested on MacOS with Intel architecture which has no leak sanitizer and address sanitizer seems broken, so I had to disable all sanitizers. The examples seem to work, but I get lots of noise in the output, not sure whom to blame.

Sorry for the big chunk of code.
I think it's ready, but feel free to review and complain :-)

PR #429, opened 2024-10-23
Performance and safety improvements

mikucionisaau:
- [x] replace `std::regex` with `std:string_view` scanning (thousands of times faster)
- [x] replaced `std::tuple` with dedicated `Version` class with named members
- [x] added virtual destructor to `backend_interface` for safety
- [x] replaced value passing with `const &` in `fplot` and more
- [ ] `clang_tidy` suggests making lots of `const` methods and more...


mikucionisaau:
Rebased and tested with GCC/Debian/testing, Windows Visual Studio Community 2022 and AppleClang/MacOS.
Feels faster and snappier.

Potential reservations: changes the API subtly, not sure about consequences for other projects which use this.

source/matplot/backend/backend_interface.h
```diff
@@ -213,7 +214,31 @@ namespace matplot {
             /// This is useful when tracing the gnuplot commands
             /// and when generating a gnuplot file.
             virtual void include_comment(const std::string &text);
-        };
+        }; // class backend_interface
+
+        /// Captures (backend) version information.
+        struct Version {
```
alandefreitas, line 220: This needs MATPLOT_EXPORTS if it's part of the API.

alandefreitas: It would also be good to maintain the snake_case convention. `version` or maybe `gnuplot::version` since there's no other backend.

source/matplot/backend/gnuplot.cpp
```diff
@@ -317,13 +317,28 @@ namespace matplot::backend {
...
+    }
+
     std::string gnuplot::default_terminal_type() {
         static std::string terminal_type;
         const bool dont_know_term_type = terminal_type.empty();
         if (dont_know_term_type) {
             terminal_type = run_and_get_output("gnuplot -e \"show terminal\" 2>&1");
-            terminal_type = std::regex_replace(terminal_type,
-                    std::regex("[^]*terminal type is ([^ ]+)[^]*"), "$1");
+            terminal_type = word_after(terminal_type, "terminal type is ");
```
alandefreitas, line 341: Heh... Great! :)

source/matplot/backend/gnuplot.cpp
```diff
@@ -334,50 +349,24 @@ namespace matplot::backend {
...
-                version == std::tuple<int, int, int>({0, 0, 0});
-            if (still_dont_know_gnuplot_version) {
-                // assume it's 5.2.6 by convention
-                version = std::tuple<int, int, int>({5, 2, 6});
+    Version gnuplot::gnuplot_version() {
+        static auto version = Version{};
+        if (!version) { // unknown version
+            const auto version_str = run_and_get_output("gnuplot --version 2>&1");
+            const auto major = word_after(version_str, "gnuplot");
+            const auto minor = word_after(major, ".");
```
alandefreitas, line 361: This pattern is a little confusing when the function is called repeatedly. Let's say the output is `"... gnuplot 5.2 patchlevel 6"`. Wouldn't `major` just be `"5"`? If so, then there's no word after `"."` because `major` doesn't include everything after `5`. That is, major is `"5"` and not `"5.2 patchlevel 6"`. On the other hand, if `major` does contain `"5.2 patchlevel 6"`, then the function name and return value are a little misleading.

source/matplot/backend/gnuplot.h
```diff
@@ -47,7 +47,7 @@ namespace matplot::backend {
         /// Identify the default terminal type in the system
         static std::string default_terminal_type();
         static bool terminal_is_available(std::string_view);
-        static std::tuple<int, int, int> gnuplot_version();
+        static Version gnuplot_version();
```
alandefreitas, line 50: So converting the output to a `tuple` wouldn't break the API? A `Version` is a better decision than a `tuple,` but it's hard to tell how many people are counting on this API.

source/matplot/core/axes_type.cpp
```diff
@@ -426,7 +426,7 @@ namespace matplot {
             run_command("set polar");
         }
         auto set_or_unset_axis = [this](class axis_type &ax,
-                                        std::string axis_name,
+                                        const std::string &axis_name,
```
alandefreitas, line 429: If that was your concern about breaking the API, both `std::string axis_name` and `std::string const& axis_name` are OK. In the implementation the `std::string axis_name` can be moved and the `std::string const& axis_name` can be copied. As far as the user is concerned, that doesn't functionally break the API. Only `std::string &axis_name` would be a problem.

source/matplot/backend/backend_interface.h
```diff
@@ -213,7 +214,31 @@ namespace matplot {
...
+            int major{0}, minor{0}, patch{0};
+            constexpr bool operator==(const Version &o) const {
+                return major == o.major && minor == o.minor && patch == o.patch;
+            }
+            constexpr bool operator!=(const Version &other) const { return !(*this == other); }
+            constexpr bool operator<(const Version &other) const {
+                if (major < other.major)
```
alandefreitas, line 227: Is it the same as `std::tie(major, minor, patch) < std::tie(other.major, other.minor, other.patch)`?

alandefreitas: Yes. Looking at the rest of the code and considering `tuple` defines `operator<`, I do prefer `Version` over `tuple` here if we were coming up with the original API now, but I don't think breaking it now brings a lot of value compared to the potential hassle of people opening issues because the API changed.

source/matplot/core/axes_type.cpp
```diff
@@ -2745,9 +2743,9 @@ namespace matplot {
     }
 
     std::vector<function_line_handle>
-    axes_type::fplot(std::vector<function_line::function_type> equations,
-                     std::array<double, 2> x_range,
-                     std::vector<std::string> line_specs) {
+    axes_type::fplot(const std::vector<function_line::function_type> &equations,
```
alandefreitas, line 2746: Same here. All these API changes are OK.

alandefreitas (commented):
Nice. Removing std::regex is refreshing. :)

source/matplot/backend/backend_interface.h
mikucionisaau: I'll make it snake case, sorry.
I could also move Version to gnuplot, I just thought the class is more general than just gnuplot.

alandefreitas: Yes. Outside `gnuplot` is also OK. I was thinking about the name-grabbing, but that's also OK. We won't need `matplot::version` for anything else.

source/matplot/core/axes_type.cpp
mikucionisaau: No concerns here, that just a lambda expression - internal stuff.
And yes, it'll bind to the same calls as previously, we just need to make sure that the lambda expression itself does not modify the parameters, but that is checked by the compiler.

source/matplot/core/axes_type.cpp
mikucionisaau: It's OK if the thing is recompiled, but I am not sure at ABI level, and I do not know if anyone cares.

source/matplot/backend/gnuplot.cpp
mikucionisaau: Right, the name `major` is misleading, I have renamed it to `major_minor` and added example in comments -- I think it's clearer when an example is upfront.

source/matplot/backend/gnuplot.h
mikucionisaau: Right, I have reverted `Version` to `std::tuple` as I did not know that one can compare them in such elegant way. I used to kill it if the code is not about generic programming (Scott Meyers book).
I learned something today, thanks!

source/matplot/backend/backend_interface.h
mikucionisaau: This is gone, sorry my bad.

mikucionisaau:
I have removed my `version` mess as `tuple` is simpler (it aleady has all operators) and thus API is preserved. 
I moved the code concerned with versioning into gnuplot class, which makes the code cleaner as the feature knowledge is maintained in one place.

source/matplot/core/figure_type.cpp
```diff
@@ -633,9 +633,7 @@ namespace matplot {
     void figure_type::run_window_color_command() {
         // In gnuplot 5.5 we have the wall function to set the axes color
         // with a rectangle workaround, which does not work well for 3d.
-        static const auto v = backend::gnuplot::gnuplot_version();
-        const bool has_wall_option =
-            std::get<0>(v) > 5 || (std::get<0>(v) == 5 && std::get<1>(v) >= 5);
```
mikucionisaau, line 638: this was my trigger :-)

Please-just-dont:
I noticed you made a change to take an argument by reference, but I've noticed that the first arguments to a lot of these functions, the first set of axes is always taken by copy, and I'm wondering why it's not taken by const reference.

mikucionisaau:
I just realized that my build directory is of 21G in size, so that's a big effort just to compile and testing is complicated by the wait (ctest does not allow interaction).
So briefly: I've got it all built with `-DMATPLOTPP_BUILD_EXAMPLES=ON -DMATPLOTPP_BUILD_TESTS=ON` and tested using `GNUTERM=dumb ninja test` and holding `ENTER` key just to complete all 545 tests, and it all passed.

Let me know if there is anything more I can do to test at my end, before dispatching this massive test on CI.

PR #461, opened 2025-05-26
Add the option to make the legend transparent

blaiselcq:
This is useful when automatically generating a large number of plots, when the best legend position is not always the same.


alandefreitas:
LGTM

Issue #362, opened 2023-07-05
Error building on linux gcc-12 


suhasghorp:
**Bug category**

- [ x] *bug - compilation error*
- [ ] *bug - compilation warning*
- [ ] *bug - runtime error*
- [ ] *bug - runtime warning*
- [ ] *bug - logic error*

**Describe the bug**

Can not build the library
**Steps to Reproduce**

```bash
# Your steps go here
mkdir build
cd build
cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="-O2" -DMATPLOTPP_BUILD_EXAMPLES=OFF -DMATPLOTPP_BUILD_TESTS=OFF
sudo cmake --build . --parallel 2 --config Release
```

**Output**

(base) sghorp@LAPTOP-NPL8D370:~/matplotplusplus/build$ cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS="-O2" -DMATPLOTPP_BUILD_EXAMPLES=OFF -DMATPLOTPP_BUILD_TESTS=OFF
Setting global options
Looking for -fsanitize=address
Adding -fsanitize=address
Looking for -fsanitize=leak
Adding -fsanitize=leak
Looking for -fsanitize=undefined
Adding -fsanitize=undefined
-- Could NOT find TIFF (missing: TIFF_LIBRARY TIFF_INCLUDE_DIR)
-- Could NOT find PNG (missing: PNG_LIBRARY PNG_PNG_INCLUDE_DIR)
-- Could NOT find FFTW (missing: FFTW_INCLUDE_DIRS)
Setting matplotplusplus compiler options
CMAKE_INSTALL_PREFIX=/usr/local
CMAKE_INSTALL_LIBDIR=lib
-- Configuring done
-- Generating done
-- Build files have been written to: /home/sghorp/matplotplusplus/build

(base) sghorp@LAPTOP-NPL8D370:~/matplotplusplus/build$ sudo cmake --build . --parallel 2 --config Release
Consolidate compiler generated dependencies of target nodesoup
[ 13%] Built target nodesoup
Consolidate compiler generated dependencies of target matplot
[ 18%] Building CXX object source/matplot/CMakeFiles/matplot.dir/backend/gnuplot.cpp.o
[ 18%] Building CXX object source/matplot/CMakeFiles/matplot.dir/backend/backend_interface.cpp.o
[...]

PR #472, opened 2025-10-14
Fix bug: qt terminal launches unexpectedly when saving plot (#362)

kunmonster:
As described in #362, when the user only wants to save the plot, the terminal is still set to qt and the graphic view is invoked.
This behavior should not occur when the plot is only exported without interactive display.

Root cause
- The gnuplot pipe is opened too early during the backend construction. https://github.com/alandefreitas/matplotplusplus/blob/master/source/matplot/backend/gnuplot.cpp#L73-L87
- Each component is rendered twice:
   1. once when it is touched
   2. once again when rendering the entire plot.
  
Fix
- Open the gnuplot pipe lazily only when show() or save() is called.
- Render the plot only when necessary (either show or save), avoiding redundant rendering.

### DIFF
--- a/source/matplot/backend/gnuplot.cpp
+++ b/source/matplot/backend/gnuplot.cpp
@@ -87,6 +87,10 @@ namespace matplot::backend {
     }
 
     void gnuplot::lazy_init_pipe() {
+        if(pipe_.opened()){
+            flush_commands();
+            return;
+        }
         int perr;
         if constexpr (windows_should_persist_by_default) {
             perr = pipe_.open("gnuplot --persist");
--- a/source/matplot/backend/gnuplot.cpp
+++ b/source/matplot/backend/gnuplot.cpp
@@ -4,13 +4,13 @@
 
 #include "gnuplot.h"
 
+#include <charconv>
+#include <cstdlib>
+#include <cstring>
+#include <iostream>
 #include <matplot/util/common.h>
 #include <matplot/util/popen.h>
-#include <iostream>
-#include <charconv>
 #include <thread>
-#include <cstring>
-#include <cstdlib>
 
 #ifdef __has_include
 #if __has_include(<filesystem>)
@@ -62,6 +62,12 @@ namespace matplot::backend {
             // 2nd option: wxt on windows, even if not default
             terminal_ = "wxt";
 #endif
+            /***
+             * TODO:when we construct gnuplot object,
+             * we should not set terminal to dumb directly,
+             * we should check the saving command
+             ***/
+
         } else if (terminal_is_available("qt")) {
             // 3rd option: qt
             terminal_ = "qt";
@@ -71,20 +77,25 @@ namespace matplot::backend {
         }
 
         // Open the gnuplot pipe_
-        int perr;
-        if constexpr (windows_should_persist_by_default) {
-            perr = pipe_.open("gnuplot --persist");
-        } else {
-            perr = pipe_.open("gnuplot");
-        }
+        // int perr;
+        // if constexpr (windows_should_persist_by_default) {
+        //     perr = pipe_.open("gnuplot --persist");
+        // } else {
+        //     perr = pipe_.open("gnuplot");
+        // }
 
-        // Check if everything is OK
-        if (perr != 0 || !pipe_.opened()) {
-            std::cerr << "Opening the gnuplot failed: ";
-            std::cerr << pipe_.error() << std::endl;
-            std::cerr << "Please install gnuplot 5.2.6+: http://www.gnuplot.info"
-                << std::endl;
-        }
+        // // if (perr == 0 && pipe_.opened()) {
+        // //     run_command("set terminal dumb");
+        // //     run_command("set output '/dev/null'");
+        // // }
+        // // Check if everything is OK
+        // if (perr != 0 || !pipe_.opened()) {
+        //     std::cerr << "Opening the gnuplot failed: ";
+        //     std::cerr << pipe_.error() << std::endl;
+        //     std::cerr
+        //         << "Please install gnuplot 5.2.6+: http://www.gnuplot.info"
+        //         << std::endl;
+        // }
     }
 
     gnuplot::~gnuplot() {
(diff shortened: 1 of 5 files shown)


source/matplot/backend/gnuplot.cpp
```diff
@@ -101,6 +86,22 @@ namespace matplot::backend {
         flush_commands();
     }
 
+    void gnuplot::lazy_init_pipe() {
+        int perr;
+        if constexpr (windows_should_persist_by_default) {
+            perr = pipe_.open("gnuplot --persist");
```
alandefreitas, line 96: So it's not only opening it later but also calling it every time we show the figure?

source/matplot/core/figure_type.cpp
```diff
@@ -131,7 +134,10 @@ namespace matplot {
         }
     }
 
-    void figure_type::show() { backend_->show(this); }
+    void figure_type::show() { 
+        // open the gnu pipe when show method is called
+        backend_->lazy_init_pipe();
```
alandefreitas: If this is going to be called every time we call "show", it's it easier to conceptually make it part of the gnuplot backend only and call it at the beginning of the show function?

source/matplot/backend/gnuplot.cpp
kunmonster: Sorry, maybe I should check whether the pipe is already open before opening it again, to avoid reopening it when calling show() or save() multiple times.

kunmonster:
Actually, I think the best way to fix the problem is that we shouldn't run the command while adding the child, instead just run the commands once in the show or save.

source/matplot/core/figure_type.cpp
kunmonster: Yep, but there isn’t an API implementation for showing in gnuplot.cpp. We just use flushcommand() to flush the buffer and display the plot. However, you’re right -- it might be more reasonable to initialize the pipe in backend_interface instead of figure_type.

source/matplot/backend/gnuplot.cpp
alandefreitas: Yes. True. If the point of the PR is only to make it lazy, that makes more sense.

But I was checking because I'm not so familiar with the changes.

source/matplot/core/figure_type.cpp
alandefreitas: Sorry. I'm a little confused. The very next line does `backend_->show(this);`. Isn't this calling "an API implementation for showing in gnuplot.cpp"? Do you mean the gnuplot backend doesn't override `show()`?

alandefreitas:
> Actually, I think the best way to fix the problem is that we shouldn't run the command while adding the child, instead just run the commands once in the show or save.

I agree. This is what called "quiet mode" or something in this implementation. In retrospect, this should be the only mode possible. But it's too late to change that unless there's a matplot++ v2.

source/matplot/core/figure_type.cpp
kunmonster: That's the truth what I want to say. But, it seems that the backend_interface::show doesn't need to be overrided. It's unnecessary.

source/matplot/backend/gnuplot.cpp
kunmonster: There's one point, this change will revise the render haviour. For example, you can run the examples/data_distribution/hisrogram14, the original code will draw step by step. But the changed code will render once during the whole time.

source/matplot/core/figure_type.cpp
alandefreitas: OK. So if I understand you correctly, `matplot::backend::gnuplot` doesn't define `matplot::backend::gnuplot::show()`, so that means `matplot::backend::gnuplot::show()` is equivalent to `matplot::backend::backend_interface::show()`.

If that's correct, we could override and define `matplot::backend::gnuplot::show()` so that 1) it calls `lazy_init_pipe()` and only then 2) calls `matplot::backend::backend_interface::show()`. The advantage of this would be that the gnuplot fix would be limited to the scope of the gnuplot backend, and figure_type would make reference to a property that doesn't belong to all backends.

source/matplot/backend/gnuplot.cpp
alandefreitas: So this means it'll only render the first time and never update it?

kunmonster: The code of histogram14 call the draw method and it will show the subplpt,but the new code will just show after calling the backed:show(). So it will just render one time.

source/matplot/core/figure_type.cpp
kunmonster: Yes, that's a nice strategy.

source/matplot/backend/gnuplot.cpp
alandefreitas: OK. Thank you for clarifying. So is that by design after this PR?

kunmonster: > OK. Thank you for clarifying. So is that by design after this PR?

No, It belongs the version before the pr. 

I mean, when you were designing this software, did you take into account the need to display one subplot at a time? Because after this change, calling show will display all the plots at once.

alandefreitas: Sorry. That's what I understood so far, please correct me if I'm wrong:

- Before this PR, draw() updates/renders the plot (especially important in "quiet" mode) and show() would do the same and also block
- After this PR, draw() wouldn't do anything until you call show() and show() would have the same expected behavior

Is my understanding correct? 

I'm sorry to be insisting. I don't want to block a PR because I don't understand it. But also can't merge a PR I don't understand clearly.

> did you take into account the need to display one subplot at a time?

Yes. That should be possible. To only one subplot at a time, but one thing at a time. That's why the distinction between 'draw' and 'show' exists. That's not even my design. It comes from all other analogous libraries.

kunmonster: Yes, your understanding is correct.
But if this use case (displaying one subplot at a time) was indeed taken into account, then this PR might need to be reconsidered.
This issue seems quite tricky to fix and may require further discussion to find a proper solution.

kunmonster: if I understant the plot logic correctly, I would like to recommend to remove the render code when adding child, and it should occus just in show or draw.

alandefreitas: > then this PR might need to be reconsidered

Exactly. I think so. `draw()` should update the plot pretty much like in matplotlib.

kunmonster: OK, I will close the PR, hope you can fix this in the future.

